PR #9, opened 2015-01-14
Reduce repetitive 'comma' implementations

frewsxcv:
So I went a little nuts here, but it makes some of the from_tokens
implementations really concise and sexy.


frewsxcv:
@skj3gg Review please

skj3gg:
You're not kidding about sexy!

PR #37, opened 2018-07-13
Convert wkt types to geo types

datanel:
For conversions that fail I use a custom function `try_into_geometry` instead of custom `TryInto` trait to not generate conflicts when `TryInto` will be stable


worace:
excited to have this!

PR #43, opened 2019-10-06
Use generic type for precision in conversion

### LINKED ISSUE #33: Use generic type for precision
http://rust-num.github.io/num/num_traits/float/trait.Float.html

pjsier:
Attempt to close #33. I based this off of the conversion implementation in [geojson](https://github.com/georust/geojson/blob/master/src/conversion.rs) and only used generic types for converting to `geo_types` objects

Initially I tried to update the `From` trait implementation, but I used `Into` instead because updating `From` gave me the following error:

```
type parameter `T` must be used as the type parameter for some local type
```

If I missed something that would make `From` work instead I can make that change. Let me know if I should change that or anything else here, thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ keywords = ["geo", "geospatial", "wkt"]
 
 [dependencies]
 geo-types = {version = "0.4", optional = true}
+num-traits = "0.2"
 
 [dev-dependencies]
 criterion = { version = "0.2" }
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -34,7 +34,7 @@ impl fmt::Display for Error {
     }
 }
 
-fn create_geo_coordinate<T>(coord: &Coord) -> geo_types::Coordinate<T>
+fn create_geo_coordinate<T>(coord: &Coord<T>) -> geo_types::Coordinate<T>
 where
     T: num_traits::Float,
 {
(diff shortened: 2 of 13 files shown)


frewsxcv:
The code changes you made look good! One thing I had in mind for #33 (which I didn't articulate well) is for the `f64` types in [`Coord`](https://github.com/georust/wkt/blob/30c53e6abd1502412dc64de1405862e630f6f56c/src/types/coord.rs#L19-L25) to also utilize `num_traits::Float`. So when WKT parsing happens, the numeric values get parsed into the generic type. Does that kinda make sense?

pjsier:
Makes sense, thanks! Wasn't sure how comprehensive you were thinking the changes should be, but that works.

It looks like some `Coord` functionality would require trait implementations on the generic type for `fmt::Display` and `str::FromStr` (and potentially others). Should I just include those traits in the definitions where they're needed and default to using `num_traits::Float`, or should there be a type defined at the root that implements all the necessary traits in addition to `num_traits::Float` like `CoordinateType` does in `geo-types`?

frewsxcv:
I'm fine with either of your options! Probably not going to be much of a difference for users of this crate

pjsier:
Gotcha, I just made those changes. I ended up needing to use `std::marker::PhantomData` to give the `Tokens` struct ownership of the generic type, but I'm not sure if that was the best option here

pjsier:
@frewsxcv let me know if there’s anything I should fix here when you have a chance, thanks!

frewsxcv:
> @frewsxcv let me know if there’s anything I should fix here when you have a chance, thanks!

Nope, looks excellent! Thanks so much!

PR #47, opened 2020-12-04
Try/From<Wkt> traits for Geometry

michaelkirk:
This is a couple separate but related things...

The main new functionality is that you can call `Geometry::from(wkt)`

Along the way, I replaced the `Try/Into` impls with `Try/From`. We should prefer TryFrom where possible since it is sometimes nicer to use and also gives you the TryInto impl for free.

Correspondingly, I deprecated methods like `try_into_geometry` in favor of `geometry.try_into` which I think is more discoverable.

One thing I was curious about - I think conventionally conversions are done via move, so the caller is responsible for any cloning, but previously all our conversions were done via references and our impls were creating new points for everything. That seemed a little strange to me, so I implemented all the new ones with the move semantics that I'm used to. I did leave the existing functionality (but rephrased them as `From`) so as not to break the API.

A couple potentially controversial things I'll call out inline...

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -15,9 +15,12 @@
 extern crate geo_types;
 extern crate num_traits;
 
-use std::fmt;
 use types::*;
 use Geometry;
+use Wkt;
+
+use std::convert::{TryFrom, TryInto};
+use std::fmt;
 
 #[derive(Debug)]
 pub enum Error {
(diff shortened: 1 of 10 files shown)


src/conversion.rs
```diff
@@ -34,156 +37,286 @@ impl fmt::Display for Error {
...
-    geo_types::Coordinate {
-        x: T::from(coord.x).unwrap(),
-        y: T::from(coord.y).unwrap(),
+    type Error = Error;
+
+    fn try_from(mut wkt: Wkt<T>) -> Result<Self, Self::Error> {
+        if wkt.items.len() == 1 {
+            Self::try_from(wkt.items.pop().unwrap())
+        } else {
+            Geometry::GeometryCollection(GeometryCollection(wkt.items)).try_into()
```
michaelkirk, line 50: It seems like a WKT can have multiple Geometries in it - even outside of the context of a GeometryCollection.

If WKT is going to be convertible to a `geo_type::Geometry` we can't have that, so I've opted to wrap it in a collection if it's more than one.

Even though this `impl` is the whole reason I created this PR, I won't be offended if you'd prefer to omit this part. I think the other changes are an improvement even if this part gets dropped.

src/conversion.rs
```diff
@@ -34,156 +37,286 @@ impl fmt::Display for Error {
...
-impl<'a, T> Into<geo_types::Geometry<T>> for &'a LineString<T>
+#[deprecated(since = "0.9", note = "use `geometry.try_into()` instead")]
+pub fn try_into_geometry<T>(geometry: &Geometry<T>) -> Result<geo_types::Geometry<T>, Error>
+where
+    T: num_traits::Float,
+{
+    geometry.clone().try_into()
+}
+
+impl<'a, T> From<&'a LineString<T>> for geo_types::Geometry<T>
```
michaelkirk, line 89: e.g. this extends the API offered by the previous `impl<'a, T> Into<geo_types::Geometry<T>> for &'a LineString<T>`

src/conversion.rs
```diff
@@ -34,156 +37,286 @@ impl fmt::Display for Error {
...
+
+    fn try_from(multi_point: MultiPoint<T>) -> Result<Self, Self::Error> {
+        let points: Vec<geo_types::Point<T>> = multi_point
             .0
-            .iter()
-            .filter_map(|p| p.0.as_ref())
-            .map(|c| create_geo_coordinate(c))
-            .collect();
+            .into_iter()
+            .map(geo_types::Point::try_from)
```
michaelkirk, line 180: 🚨 This is a change in behavior. Previously, though we Err when converting an empty Point, we'd silently drop empty Point's in a MultiPoint. That seems surprising to me, but if it was intentional, I can put it back.

frewsxcv: it was not intentional. your change looks good 👍

frewsxcv (approved):

michaelkirk:
bors r=frewsxcv

michaelkirk:
😭

frewsxcv:
approved! try again?

michaelkirk:
bors retry

michaelkirk:
bors r=frewsxcv

michaelkirk:
> approved! try again?

Thanks @frewsxcv !

PR #52, opened 2021-01-07
use TryFrom universally

michaelkirk:
---

Deprecated custom conversion functions and replaced them with conventional `From`/`TryFrom` implementations.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -15,9 +15,12 @@
 extern crate geo_types;
 extern crate num_traits;
 
-use std::fmt;
 use types::*;
 use Geometry;
+use Wkt;
+
+use std::convert::{TryFrom, TryInto};
+use std::fmt;
 
 #[derive(Debug)]
 pub enum Error {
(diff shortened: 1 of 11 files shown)


src/conversion.rs
```diff
@@ -34,156 +37,286 @@ impl fmt::Display for Error {
...
-impl<'a, T> Into<geo_types::Geometry<T>> for &'a Polygon<T>
+impl<'a, T> TryFrom<&'a MultiPoint<T>> for geo_types::Geometry<T>
 where
     T: num_traits::Float,
 {
-    fn into(self) -> geo_types::Geometry<T> {
-        w_polygon_to_g_polygon(self).into()
+    type Error = Error;
+
+    fn try_from(multi_point: &'a MultiPoint<T>) -> Result<Self, Self::Error> {
```
michaelkirk, line 165: I've inverted the existing `Into` implements to be `From` which provides `A::from(b)` while implicitly generating a `b.into(a)` to maintain existing functionality.

In general I'm not a fan of the implicit cloning that happens with this kind of type signature, which returns an owned geometry from a refernce to a wkt, but I wasn't feeling bold enough to break the existing API.

src/conversion.rs
```diff
@@ -34,156 +37,286 @@ impl fmt::Display for Error {
...
+            .into_iter()
+            .map(geo_types::Polygon::from)
+            .collect();
 
-        geo_multipolygon.into()
+        geo_types::MultiPolygon(geo_polygons)
     }
 }
 
+#[deprecated(since = "0.8.1", note = "use `geometry_collection.try_into()` instead")]
```
michaelkirk, line 211: Since we're pre v1, I prefer to only bump the patch version, even when adding new functionality to avoid locking people to an unnecessarily old version. 

As a side note, I'd love to start v1'ing our crates.

michaelkirk:
Lol I'm losing my mind and already merged this in https://github.com/georust/wkt/pull/47

PR #53, opened 2021-01-14
Mkirk/geo types 0.7

michaelkirk:
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 keywords = ["geo", "geospatial", "wkt"]
 
 [dependencies]
-geo-types = {version = "0.6", optional = true}
+geo-types = {version = "0.7", optional = true}
 num-traits = "0.2"
 
 [dev-dependencies]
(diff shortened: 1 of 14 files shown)


src/conversion.rs
```diff
@@ -12,16 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-extern crate geo_types;
-extern crate num_traits;
```
michaelkirk, line 16: I did some import tidying, no change in behavior from any of that.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ readme = "README.md"
 keywords = ["geo", "geospatial", "wkt"]
 
 [dependencies]
-geo-types = {version = "0.6", optional = true}
+geo-types = {version = "0.7", optional = true}
 num-traits = "0.2"
```
frewsxcv, line 14: is num_traits unused now?

frewsxcv (approved):

Cargo.toml
frewsxcv: nvm, forgot that the wkt structures are generic over num_traits::Float

PR #54, opened 2021-01-14
Mkirk/debug

michaelkirk:
---

Based on #53, so review that first.

I'm not actually sure if we should do this. Or if we should maybe go even farther and have "WktFloat: Display  + FromStr + Default" as well. Since we're bumping API anyway, thought it might be a good time to sneak it in.

It's not clear to me how useful the crate would be for numeric types that didn't implement all of these things.


frewsxcv (approved):

michaelkirk:
bors r=frewsxcv

Issue #49, opened 2020-12-04
displayable errors: the trait `StdError` is not implemented for `wkt::conversion::Error`


michaelkirk:
Some of the other georust crates use thiserror. Is that something we should use here?

frewsxcv:
no strong opinions from me between std::error::Error and thiserror! either would be an improvement

PR #57, opened 2021-01-30
specific errors and `try_into` for inner geo-types

michaelkirk:
---

**DRAFT**: depends on https://github.com/georust/geo/pull/614 being released first.

Add TryFrom for converting directly to `geo_types::Geometry` enum members. 

```
let ls: geo_types::LineString = geo_types::LineString::try_from(wkt).unwrap();
```

Previously it was:
```
let geom: geo_types::Geometry = geo_types::Geometry::from(wkt).unwrap();
let ls: geo_types::LineString = geo_types::LineString::try_from(geom).unwrap();
```

This introduced two new error cases, so I've also introduced `thiserror` which solves #49

FIXES https://github.com/georust/wkt/issues/49

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -17,23 +17,23 @@ use Geometry;
 use Wkt;
 
 use std::convert::{TryFrom, TryInto};
-use std::fmt;
 
 use geo_types::CoordFloat;
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum Error {
+    #[error("The WKT Point was empty, but geo_type::Points cannot be empty")]
     PointConversionError,
-}
-
-impl fmt::Display for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            Error::PointConversionError => {
-                f.write_str("impossible to convert empty point to geo_type point")
-            }
-        }
-    }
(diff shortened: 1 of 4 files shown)


src/conversion.rs
```diff
@@ -51,6 +51,50 @@ where
...
+                            let item = wkt.items.pop().unwrap();
+                            let geometry = geo_types::Geometry::try_from(item)?;
+                            Self::try_from(geometry).map_err(|e| {
+                                match e {
+                                    geo_types::Error::MismatchedGeometry { expected, found } => {
+                                        Error::MismatchedGeometry { expected, found }
+                                    }
+                                    // currently only one error type in geo-types error enum, but that seems likely to change
+                                    #[allow(unreachable_patterns)]
+                                    other => Error::External(Box::new(other)),
```
michaelkirk, line 74: I thought it was worth future-proofing, so we don't have to bump WKT's minimum version of geo-types again if we add to the error enum.

urschrei:
This is a really big ergonomic improvement; the enum is really useful, but it's been historically awkward to work with and this removes a lot of that friction 👍👍👍

michaelkirk:
Ok - updated now that the requisite geo-type 0.7.1 has been published.

Are you OK with merging this @urschrei?

urschrei:
Yep, lgtm!

michaelkirk:
bors r=urschrei

Excellent thanks!

PR #59, opened 2021-02-12
impl serde::Deserialize for Wkt and Geometry (and helper function)

woshilapin:
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,12 +10,14 @@ readme = "README.md"
 keywords = ["geo", "geospatial", "wkt"]
 
 [dependencies]
-geo-types = {version = "0.7.1", optional = true}
+geo-types = { version = "0.7.1", optional = true }
 num-traits = "0.2"
+serde = { version = "1.0", default-features = false, optional = true }
 thiserror = "1.0.23"
 
 [dev-dependencies]
 criterion = { version = "0.2" }
+serde = { version = "1.0", default-features = false, features = ["derive"] }
 
 [features]
 default = ["geo-types"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ thiserror = "1.0.23"
 [dev-dependencies]
 criterion = { version = "0.2" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde_json = "1.0"
 
 [features]
 default = ["geo-types"]
--- a/src/deserialize.rs
+++ b/src/deserialize.rs
@@ -1,0 +1,170 @@
+use crate::{Geometry, Wkt, WktFloat};
+use serde::de::{Deserializer, Error, Visitor};
+use std::{
+    default::Default,
+    fmt::{self, Debug},
+    marker::PhantomData,
+    str::FromStr,
+};
+
+struct WktVisitor<T> {
+    _marker: PhantomData<T>,
+}
+
+impl<T> Default for WktVisitor<T> {
+    fn default() -> Self {
+        WktVisitor {
+            _marker: PhantomData::default(),
+        }
+    }
+}
+
+impl<'de, T> Visitor<'de> for WktVisitor<T>
+where
+    T: FromStr + Default + Debug + WktFloat,
+{
+    type Value = Wkt<T>;
+    fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(formatter, "a valid WKT format")
+    }
+    fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        Wkt::from_str(s).map_err(|e| serde::de::Error::custom(e))
+    }
+}
+
+impl<'de, T> serde::Deserialize<'de> for Wkt<T>
+where
+    T: FromStr + Default + Debug + WktFloat,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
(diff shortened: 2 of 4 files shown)


michaelkirk (approved):
This looks great to me, thank you! 

I'm going to leave it open for a few days before merging in case anyone else would like to review.

michaelkirk:
bors try

src/deserialize.rs
```diff
@@ -0,0 +1,201 @@
...
+/// This is a helper function to convert directly from WKT format into a geo_types::Geometry.
+/// ```
+/// # extern crate wkt;
+/// # extern crate geo_types;
+/// # extern crate serde_json;
+/// use geo_types::Geometry;
+///
+/// #[derive(serde::Deserialize)]
+/// struct MyType {
+///     #[serde(deserialize_with = "wkt::deserialize::deserialize_wkt_geometry")]
```
antoine-de, line 106: maybe we can export the module to only do `wkt::deserialize_geometry`?

src/deserialize.rs
```diff
@@ -0,0 +1,201 @@
...
+pub fn deserialize_wkt_geometry<'de, D>(
+    deserializer: D,
+) -> Result<geo_types::Geometry<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    use serde::Deserialize;
+    use std::convert::TryInto;
+    let geometry: Geometry<f64> = Geometry::deserialize(deserializer)?;
+    let geometry: geo_types::Geometry<f64> = geometry.try_into().unwrap();
```
antoine-de, line 124: can't we export an error here?

*very* minor: wouldn't it be better to `use std::convert::TryInto;` just before using it?

michaelkirk: I agree we don't want to `unwrap` here, thanks for noticing @antoine-de

src/deserialize.rs
woshilapin: Yes, I thought about that, glad you proposed it. Done in f5e5d359f76fe0aa25767e4f766e61d42b651bbf.

src/deserialize.rs
woshilapin: I completely forgot that `.unwrap()`, oops. Fixed in d5a8de371219c84d9fff67a8e501d44f5c397a1d. I also reorganized the code a bit with a `.and_then()`, but please tell if you prefer the previous version, I can definitely change it back.

woshilapin: I also made the helper function generic over the type `T` of `Geometry<T>` because the first version was useful for `Geometry<f64>` but was useless for `Geometry<f32>`. See commit be202f413bfc1f2dbdfff3d0333cb8c0606025f0.

michaelkirk: These changes all look good to me, thank you!

michaelkirk:
bors try

src/deserialize.rs
```diff
@@ -0,0 +1,200 @@
...
+/// This is a helper function to convert directly from WKT format into a geo_types::Geometry.
+/// ```
+/// # extern crate wkt;
+/// # extern crate geo_types;
+/// # extern crate serde_json;
+/// use geo_types::Geometry;
+///
+/// #[derive(serde::Deserialize)]
+/// struct MyType {
+///     #[serde(deserialize_with = "wkt::deserialize_geometry")]
```
michaelkirk, line 106: oh right, there is a special dance to do when a doc-test requires a feature flag.

We have a similar example here in `proj` which requires the optional `geo-types` feature:

https://github.com/georust/proj/blob/master/src/lib.rs#L175

michaelkirk: Actually, forget that. This whole module is only included conditionally with the `serde` feature. I'm not immediately sure what to make of the CI failure. I can look at it further next week.

https://github.com/georust/wkt/runs/1891796241

src/lib.rs
```diff
@@ -43,6 +43,13 @@ pub use towkt::ToWkt;
 #[cfg(feature = "geo-types")]
 pub mod conversion;
 
+#[cfg(feature = "serde")]
+extern crate serde;
+#[cfg(feature = "serde")]
+pub mod deserialize;
+#[cfg(feature = "serde")]
```
michaelkirk, line 50: ah! it's failing because we need both serde and geo-types for this method.

woshilapin: Ah right, I failed this one. Nice CI setup to catch that by the way. Fixed in 2c6f669686dc334ada95442672a7a52418e85df9.

antoine-de (approved):

Cargo.toml
```diff
@@ -10,15 +10,18 @@ readme = "README.md"
...
 thiserror = "1.0.23"
 
 [dev-dependencies]
 criterion = { version = "0.2" }
+serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde_json = "1.0"
 
 [features]
-default = ["geo-types"]
+default = ["geo-types", "serde"]
```
frewsxcv, line 24: From what I've seen in the ecosystem, Serde support is usually opt-in and not included as a default feature. If I had to guess, I'd say more often then not, users won't need Serde support when using this crate. Would love to hear from others

michaelkirk: Yeah, I think you're right. It should probably not be included in `default`.

woshilapin: Oops, this was definitely not intended as I used it only locally to facilitate the development. Good catch, thank you. I amended the last commit.

frewsxcv (approved):

michaelkirk (approved):
bors r+

Ok, I think this has gotten enough feedback. Thanks @woshilapin!

PR #60, opened 2021-02-15
add docs/crates badge. remove deprecated CI badge

michaelkirk:
---

Added a docs/crates badge.

🔥hot take🔥:  I removed the ci badge all together.
we no longer use travis ci. I could have changed this to use a GH actions
badge, but people can already see the build status next to the commit on
the front page. I don't think it warrants a redundant badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,15 @@
 wkt
 ===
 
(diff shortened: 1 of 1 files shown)


frewsxcv (approved):
Generally not a fan of badges in general, though I know others like them so I don't usually mind if someone adds them. Thanks for noticing and updating the badges to be more relevant

Issue #61, opened 2021-02-18
`POINT EMPTY` vis-à-vis `geo_types`


michaelkirk:
One point of friction when  mapping WKT to geo_types is `POINT EMPTY`.

Similar EMPTY types present no such problem: `GEOMETRYCOLLECTION  EMPTY`, `LINESTRING EMPTY`, `POLYGON EMPTY` all have intuitive corollaries in geo_types. (`GeometryCollection(vec![])`, `LineString(vec![])`, `Polygon { exterior: vec![], interiors: vec![] }`

You might say:

> Obviously `POINT EMPTY` maps to `Option::<geo_types::Point>::None`!

To which I say:

> `GEOMETRYCOLLECTION(POINT EMPTY)`
or
> `GEOMETRYCOLLECTION(GEOMETRYCOLLECTION(POINT EMPTY))`
> `GEOMETRYCOLLECTION(POINT 1 1, GEOMETRYCOLLECTION(POINT EMPTY))`

The conversion of which is not obvious to me.

A few options:

1. Do nothing. Whenever converting to geo-types, continue to error whenever we encounter a `POINT EMPTY`, force the caller to reconcile any confusion. This is great because we wash our hands of the problem, but I think we'd be doing a disservice to, as policy, force every user to re-invent this essential wheel from scratch.

2. add a `deserialize_point -> Option<geo_types::Point>` method to complement `deserialize_geometry`. 
```
#[cfg(feature = "geo-types")]
pub fn deserialize_point<'de, D, T>(deserializer: D) -> Result<Option<geo_types::Point<T>>, D::Error>
    where
        D: Deserializer<'de>,
        T: FromStr + Default + WktFloat,
{
    use serde::Deserialize;
    Wkt::deserialize(deserializer).and_then(|wkt: Wkt<T>| {
        use std::convert::TryFrom;
        geo_types::Point::try_from(wkt).map(|p| Some(p))
            .or_else(|e| {
            if let crate::conversion::Error::PointConversionError = e {
                // map a WKT: 'POINT EMPTY' to an `Option<geo_types::Point>::None`
                return Ok(None);
            }

            Err(D::Error::custom(e))
        })
    })
}
```

This would *only* solve the top level `POINT EMPTY` case, and would continue to error on `GEOMETRYCOLLECTION(POINT EMPTY)` case. Would this half measure be worthwhile or only amplify the confusion?

[...]

PR #62, opened 2021-02-20
deserialize point

michaelkirk:
  -  > Add helper function for deserializing from WKT format into geo_types::Geometry
    
      I think the existing unreleased note pretty much covers it 
---

Addresses one little bit of https://github.com/georust/wkt/issues/61, though leaves unanswered the bigger questions, around how to handle `GEOMETRYCOLLETION(POINT EMPTY)` etc.

This might be controversial as it arguably muddies the water for addressing #61 more comprehensively. Please take a look at #61 for context.

### DIFF
--- a/src/deserialize.rs
+++ b/src/deserialize.rs
@@ -130,7 +130,7 @@ where
 /// # extern crate wkt;
 /// # extern crate geo_types;
 /// # extern crate serde_json;
-/// use geo_types::Geometry;
+/// use geo_types::Point;
 ///
 /// #[derive(serde::Deserialize)]
 /// struct MyType {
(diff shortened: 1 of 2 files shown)


src/deserialize.rs
```diff
@@ -124,6 +124,49 @@ where
     })
 }
 
+/// Deserializes directly from WKT format into a `Option<geo_types::Point>`.
```
frewsxcv, line 127:
```suggestion
/// Deserializes directly from WKT format into a `Option<geo_types::Point>`.
///
/// # Examples
///
```

frewsxcv (approved):

michaelkirk:
bors r=frewsxcv

PR #64, opened 2021-03-02
Support 'POINT EMPTY' conversion to geo_types

### LINKED ISSUE #61: `POINT EMPTY` vis-à-vis `geo_types`

rmanoka:
---

closes  #61

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -291,9 +291,9 @@ where
             Geometry::Point(g) => {
                 // Special case as `geo::Point` can't be empty
                 if g.0.is_some() {
-                    geo_types::Geometry::Point(g.try_into()?)
+                    geo_types::Point::try_from(g)?.into()
                 } else {
-                    geo_types::Geometry::MultiPoint(geo_types::MultiPoint(vec![]))
+                    geo_types::MultiPoint(vec![]).into()
                 }
             }
             Geometry::LineString(g) => geo_types::Geometry::LineString(g.into()),
@@ -365,8 +365,8 @@ mod tests {
 
     #[test]
     fn convert_empty_point() {
-        let point = Point(None).as_item();
-        let res: Result<geo_types::Geometry<f64>, Error> = point.try_into();
+        let point = Point(None);
+        let res: Result<geo_types::Point<f64>, Error> = point.try_into();
         assert!(res.is_err());
     }
 
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -288,7 +288,14 @@ where
 
     fn try_from(geometry: Geometry<T>) -> Result<Self, Self::Error> {
         Ok(match geometry {
-            Geometry::Point(g) => geo_types::Geometry::Point(g.try_into()?),
+            Geometry::Point(g) => {
+                // Special case as `geo::Point` can't be empty
+                if g.0.is_some() {
+                    geo_types::Geometry::Point(g.try_into()?)
+                } else {
+                    geo_types::Geometry::MultiPoint(geo_types::MultiPoint(vec![]))
+                }
+            }
             Geometry::LineString(g) => geo_types::Geometry::LineString(g.into()),
             Geometry::Polygon(g) => geo_types::Geometry::Polygon(g.into()),
             Geometry::MultiLineString(g) => geo_types::Geometry::MultiLineString(g.into()),
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@
 ## Unreleased
 
 * Add new entries here
+* Support `POINT EMPTY` in conversion to `geo_types`.
+  Converts to `MultiPoint([])`.
+  * <https://github.com/georust/wkt/pull/64>
 
 ## 0.9.1
 
--- a/benches/parse.rs
+++ b/benches/parse.rs
@@ -8,14 +8,14 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
     c.bench_function("parse small", |bencher| {
         let s = include_str!("./small.wkt");
         bencher.iter(|| {
-            let _ = wkt::Wkt::from_str(s).unwrap();
+            let _ = wkt::Wkt::<f64>::from_str(s).unwrap();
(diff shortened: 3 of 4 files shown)


lnicola (approved):

src/conversion.rs
```diff
@@ -288,7 +288,14 @@ where
 
     fn try_from(geometry: Geometry<T>) -> Result<Self, Self::Error> {
         Ok(match geometry {
-            Geometry::Point(g) => geo_types::Geometry::Point(g.try_into()?),
+            Geometry::Point(g) => {
+                // Special case as `geo::Point` can't be empty
+                if g.0.is_some() {
```
michaelkirk, line 293: Very minor thing, but we should be able to write this in an infallible way, like:
```
 if let Some(c) = g.0 {
    geo_types::Geometry::Point(geo_types::Point(c.into()))
```

michaelkirk (commented):
Excellent! Thanks for taking this one!

I see one failing test when I run locally:
```
failures:

---- conversion::tests::convert_empty_point stdout ----
thread 'conversion::tests::convert_empty_point' panicked at 'assertion failed: res.is_err()', src/conversion.rs:370:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    conversion::tests::convert_empty_point

test result: FAILED. 60 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
```

rmanoka:
> ```
> failures:
>     conversion::tests::convert_empty_point
> 
> test result: FAILED. 60 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
> ```

Good catch!  I overlooked this test.

src/conversion.rs
rmanoka: I wanted to allow any other error handling / extensions we might add in the future here.

PR #66, opened 2021-03-09
Approximately 40% faster, not all numeric chars are digits

softdevca:
- [X] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [X] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 * Add new entries here
+* Approximately 40% faster according to `cargo bench`.
+* Some "numeric" characters like `¾` and `①` were being treated as digits.
 * Support `POINT EMPTY` in conversion to `geo_types`.
   Converts to `MultiPoint([])`.
   * <https://github.com/georust/wkt/pull/64>
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -80,21 +80,13 @@ where
             ')' => Some(Token::ParenClose),
             ',' => Some(Token::Comma),
             c if is_numberlike(c) => {
-                let mut number = self.read_until_whitespace().unwrap_or_default();
-                if c != '+' {
-                    // Prepend the character because the string likely has capacity.
-                    number.insert(0, c);
-                }
+                let number = self.read_until_whitespace(if c == '+' { None } else { Some(c) });
                 match number.parse::<T>() {
                     Ok(parsed_num) => Some(Token::Number(parsed_num)),
                     Err(_) => None,
                 }
             }
-            c => {
-                let mut word = self.read_until_whitespace().unwrap_or_default();
-                word.insert(0, c);
-                Some(Token::Word(word))
-            }
+            c => Some(Token::Word(self.read_until_whitespace(Some(c)))),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/tokenizer.rs
```diff
@@ -101,40 +101,28 @@ where
...
+            match next_char {
+                '\0' | '(' | ')' | ',' => break, // Just stop on a marker
+                c if is_whitespace(c) => {
+                    let _ = self.chars.next();
+                    break;
+                }
+                _ => {
+                    let _ = self.chars.next();
+                    result
+                        .get_or_insert_with(|| String::with_capacity(16))
```
michaelkirk, line 119: Can you leave a code comment with the justification for this `16` constant?

Why do we need it at all? Why isn't it 8? or 32?

src/tokenizer.rs
```diff
@@ -85,14 +80,19 @@ where
             ')' => Some(Token::ParenClose),
             ',' => Some(Token::Comma),
             c if is_numberlike(c) => {
-                let number = c.to_string() + &self.read_until_whitespace().unwrap_or_default();
-                match number.trim_start_matches('+').parse::<T>() {
+                let mut number = self.read_until_whitespace().unwrap_or_default();
+                if c != '+' {
+                    // Prepend the character because the string likely has capacity.
+                    number.insert(0, c);
```
michaelkirk, line 86: I'm not sure how significant the cost is, but inserting at the beginning is typically slower.

Do you think it would be worth changing:
```
fn read_until_whitespace(&mut self) -> Option<String>
```

To something like:
```
fn read_until_whitespace(&mut self, string: String) -> String
```

And then :
```
let number = self.read_until_whitespace(c.to_string());
```

It might not be worth it...

src/tokenizer.rs
```diff
@@ -29,19 +29,14 @@ where
...
 }
 
+#[inline]
 fn is_numberlike(c: char) -> bool {
-    match c {
-        c if c.is_numeric() => true,
-        '.' | '-' | '+' => true,
-        _ => false,
-    }
+    c == '.' || c == '-' || c == '+' || c.is_ascii_digit()
```
michaelkirk, line 39: I think this is the only part relevant for the correctness fix - the rest of the changes are about perf, right?

michaelkirk (commented):
Hi @softdevca - thanks for the pull request. 

The changes sound good, but I had a few questions.

src/tokenizer.rs
softdevca: Yes, correct.

CHANGES.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased
 
 * Add new entries here
+* Approximately 40% faster according to `cargo bench`.
+* Some "numeric" characters like `¾` and `①` were being treated as digits.
```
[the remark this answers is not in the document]
softdevca: I cannot create an example that would be effected at this time.  The tokenizer was returning None when encountering a character that was numeric but not a digit, however the parser conveniently avoided the problem by treating the None as an end of statement. It may be fine to remove it from the changelog.

src/tokenizer.rs
softdevca: Indeed, inserting at the beginning is typically slower, however allocation is generally much, much slower than moving some characters.   At this point we know the strings are short and likely have some spare capacity.  Previously it was doing char.to_string(), which creates a string with a capacity of 1 that will immediately be re-allocated when the string returned by read_until_whitespace is appended. 

Your suggestion is worth considering.  Even if it's not faster it would keep the string building in the same place.

src/tokenizer.rs
softdevca: Comment in next commit.

michaelkirk (approved):
Excellent, this LGTM. 

I'll wait a couple days before merging in case anyone else wants to review.

PR #70, opened 2021-06-11
Relax the parens requirement for multipoint

audunska:
Postgis' `ST_AsText` function will serialize multipoints like
`MULTIPOINT(0 1, 2 3)`,
which would fail to deserialize because we expect a string like
`MULTIPOINT((0 1), (2 3))`.

In fact, the Postgis `ST_GeomFromText` function even accepts mixed
bracketing, like
`MULTIPOINT((0 1), 2 3)`.

This PR updates the parsing of multipoints to allow all these styles.

---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+* Now accepts `MULTIPOINT`s with less parentheses, as output by `ST_AsText` in postgis:
+  `MULTIPOINT(0 1, 2 3)` as opposed to `MULTIPOINT((0 1), (2 3))`
+
 ## 0.9.2 - 2020-04-30
 ### Added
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased
+### Changed
+* Now accepts `MULTIPOINT`s with less parentheses, as output by `ST_AsText` in postgis:
+  `MULTIPOINT(0 1, 2 3)` as opposed to `MULTIPOINT((0 1), (2 3))`
```
michaelkirk, line 6:
```suggestion
  `MULTIPOINT(0 1, 2 3)` in addition to `MULTIPOINT((0 1), (2 3))`
```

Might help clarify that the old format is still accepted.

michaelkirk (approved):
This LGTM, but I'll leave it open a couple days in case anyone else wants to comment.

PR #71, opened 2022-01-19
Migrate to 2021 edition and add examples

urschrei:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,8 +8,13 @@ jobs:
     container:
       image: georust/geo-ci
     steps:
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          components: rustfmt, rust-src
       - name: Checkout repository
         uses: actions/checkout@v2
-      - run: cargo install cargo-all-features
-      - run: cargo build-all-features
-      - run: cargo test-all-features
+      - run: cargo test --all-features && cargo test --no-default-features
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![doc(html_logo_url = "https://raw.githubusercontent.com/georust/meta/master/logo/logo.png")]
 // Copyright 2014-2015 The GeoRust Developers
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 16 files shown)


lnicola:
We could go right to 2021, I think?

lnicola (approved):

urschrei:
Oh, I was just being conservative because nothing else in georust is 2021 yet

lnicola:
Might be a good place to start, since `wkt` isn't tested on older compilers :smile:.

lnicola:
We need something like:

```yaml
    - name: Install Rust toolchain
      uses: actions-rs/toolchain@v1
      with:
        toolchain: stable
        profile: minimal
        override: true
        components: rustfmt, rust-src
```

urschrei:
Two last things I'd like to nail down:

- a one-line doc comment for the `Wkt` struct (what's it for?)
- a one-line doc comment for the `Geometry` enum (what's it for?)

.github/workflows/test.yml
```diff
@@ -8,8 +8,13 @@ jobs:
...
+          toolchain: stable
+          profile: minimal
+          override: true
+          components: rustfmt, rust-src
       - name: Checkout repository
         uses: actions/checkout@v2
-      - run: cargo install cargo-all-features
-      - run: cargo build-all-features
-      - run: cargo test-all-features
+      - run: cargo test
```
michaelkirk, line 20: Was it too slow to run all the feature combinations?

I think we should do at least `cargo test --all-features` and `cargo test --no-default-features`

urschrei: Good point, although `no-default-features` will cause our examples to fail…

src/lib.rs
```diff
@@ -12,24 +13,70 @@
...
+//! use wkt::Wkt;
+//! let point: Wkt<f64> = Wkt::from_str("POINT(10 20)").unwrap();
+//! ```
+//!
+//! ```
+//! // Convert to a geo_types primitive from a Wkt struct
+//! use std::convert::TryInto;
+//! use wkt::Wkt;
+//! use geo_types::Point;
+//! // point's items field contains a wkt::Geometry enum member
```
michaelkirk, line 35: Yeah... items is weird. I guess the idea is that you can express multiple geometries without explicitly putting them into a GeometryCollection? 

I tried to track down the corresponding section in the spec, but chickened out after a few minutes.

http://docs.opengeospatial.org/is/18-010r7/18-010r7.html

michaelkirk: Maybe better to not even mention items in the introductory example, and only show it in the case where there are multiple items? Otherwise is there *any* reason to even know about items?

michaelkirk: I guess without the geo-types feature, `items` is the only way to get the geometry =(

michaelkirk (approved):
Thanks for adding some WKT docs! I agree it's unnecessarily difficult to get started with the crate without them.

None of my feedback should be considered blocking.

src/lib.rs
urschrei: It would be great to know whether anyone actually uses the multiple geometry functionality in the wild…

.github/workflows/test.yml
lnicola: I think there's a way to declare required features for examples. Testing with `--no-default-features` still seems desirable.

urschrei: Yeah, trying to find the details now instead of putting `cfg`s everywhere

lnicola: https://stackoverflow.com/a/61417700 doesn't look great, though.

urschrei: Oh nooo

nyurik:
@urschrei should this be merged?  There are two approvals, but no merge, so high chance of bitrot

urschrei:
Oh, yeah. I think I was just leaving it for a few days.

PR #76, opened 2022-02-24
Use geo_types macros to future proof

nyurik:
Use `coord!` macro instead of `geo_types::Coordinate<T> { x, y }`. Doing this makes coordinate creation consistent, and allows more flexibility in the geo_types repo to introduce object changes.

---


michaelkirk:
bors r+

not sure why the previous one didn't seem to take...

PR #79, opened 2022-02-24
BREAKING: move `Wkt::from_str` to `FromStr` trait

nyurik:
Most of the user's code will stay the same, but will require `use std::str::FromStr;`

See also discussion in #77 

---


urschrei:
Bors r+

PR #85, opened 2022-03-22
wkt adapted for 3d/m change in geo

nyurik:
This PR updates WKT to the pending https://github.com/georust/geo/pull/742

Note that merging this PR might make the main branch unpublishable because it requires patched crates


nyurik:
won't be needed, closing

kylebarron:
This is now stale, and given that geo-types 3d/m did not go through, and we now have Z, M, ZM dimension support through geo-traits, we have viable workarounds anyways.

PR #87, opened 2022-04-01
Update pull_request_template.md

categulario:
---
It was useful to find the note about adding changes to the changelog. It might be as useful to ask people to run cargo fmt before making the pull-request


michaelkirk:
Thanks!

PR #86, opened 2022-03-31
impl Display for Wkt

categulario:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

to address #83 and make it easier to go from Wkt to string representation. Includes test

Issue #73, opened 2022-01-20
Remove ToWkt?


frewsxcv:
Does this get used anywhere? If so seems like we could replace usages with `Into<ToWkt>`

lnicola:
It's more discoverable than `into()`, but such are Rust APIs. I guess it's fine to remove it.

frewsxcv:
@lnicola Are you currently using `ToWkt` anywhere? I agree `Into<ToWkt>` is less discoverable, but my guess is nobody is using `ToWkt` to begin with

lnicola:
Ah, no, I'm not using it, but as a user I'd be fine with any reasonable API.

categulario:
New user here. As a user, I'd love to see (and am yet to find) an easy way to convert from geo-types to Wkt for the purposes of getting the WKT string representation.

If ye permit, I can make a PR in the next hours with such impls that would make life much easier.

michaelkirk:
It's used in a few places at least: https://github.com/search?l=&q=wkt%3A%3AToWkt+language%3ARust&type=code

Some additional uses here, but includes some false positives from identically named traits: https://github.com/search?l=&q=ToWkt+language%3ARust&type=code

Apart from arguably being more discoverable, there might be a reason to keep it as a means of including helper methods liek the `to_wkt_string()` proposed in #89

categulario:
I don't think it is actually more discoverable. For instance, it is another thing to bring into scope that you have to search the source code to know of (which I had to do).

Also, it seems like adding unrelated things to a trait that was designed to convert things to Wkt.

categulario:
I really don't think `trait ToWkt` is the right abstraction. Adding one more (actually useful) method that people will depend on will tie the existence of this trait to this library forever and prevent better abstractions to be developed

categulario:
less than 50 usages is a good number, pretty low, and a good opportunity to improve the abstractions this crate has to offer

urschrei:
I'm on the fence here. I agree that there isn't much usage (looks like three or so active crates), so deprecating it (we wouldn't be removing it immediately anyway, in line with our "try not to make dependent crate maintainers miserable" policy) isn't a major concern. I like both #88 **and** #89's new method, fwiw.

michaelkirk:
>  prevent better abstractions to be developed

Can you expound on this? Did you have something specific in mind? Are you thinking something more like the hypothetical traits you mentioned in #88: 

[...]

Issue #48, opened 2020-12-04
Document reading and writing examples


michaelkirk:
It'd be nice to have some top-level docs with examples showing some basic usage of this crate.

michaelkirk:
Along with that, we should include some examples of the serde integration once #59 is merged.

Issue #55, opened 2021-01-29
Better way to serialize to wkt?


michaelkirk:
I was trying to write a geo-type as WKT to file, and this is what I came up with:

```
    let points = include!("../src/algorithm/test_fixtures/louisiana.rs");
    let geo_linestring: geo::LineString<f32> = points.into();
    use wkt::ToWkt;
    let wkt: wkt::Wkt<_> = geo::Geometry::from(geo_linestring).to_wkt();
    let wkt_linestring = wkt.items.first().unwrap();
    let serialized = format!("{}", wkt_linestring);
    println!("{}", &serialized);
```

Is there already better way?

I was hoping for something like:
```diff
    let points = include!("../src/algorithm/test_fixtures/louisiana.rs");
    let geo_linestring: geo::LineString<f32> = points.into();
    use wkt::ToWkt;
    let wkt: wkt::Wkt<_> = geo::Geometry::from(geo_linestring).to_wkt();
-   let wkt_linestring = wkt.items.first().unwrap();
-   let serialized = format!("{}", wkt_linestring);
+   let serialized = wkt.to_string()
    println!("{}", &serialized);
```

Or like [`serde_jsons::to_writer`](https://docs.serde.rs/serde_json/fn.to_writer.html)
```diff
    let points = include!("../src/algorithm/test_fixtures/louisiana.rs");
    let geo_linestring: geo::LineString<f32> = points.into();
    use wkt::ToWkt;
    let wkt: wkt::Wkt<_> = geo::Geometry::from(geo_linestring).to_wkt();
-   let wkt_linestring = wkt.items.first().unwrap();
-   let serialized = format!("{}", wkt_linestring);
+   let file = File::create("/tmp/foo").unwrap();
+   let mut writer = BufWriter::new(file);
+   wkt::to_writer(&mut writer, wkt);
    println!("{}", &serialized);
```

Any thoughts?

michaelkirk:
Somewhat related: proposal to get rid of items array here: #72

callpraths:
I believe this will be closed by https://github.com/georust/wkt/pull/86

categulario:
I would get even better with #88

michaelkirk:
With #89 this becomes:

```diff
    let points = include!("../src/algorithm/test_fixtures/louisiana.rs");
    let geo_linestring: geo::LineString<f32> = points.into();
    use wkt::ToWkt;
[...]

PR #88, opened 2022-04-01
WIP: Impl/from geo types

categulario:
- [x] I ran cargo fmt
---

addresses #73 and #48.

This adds impls of From<Type> for every type and the Geometry enum in `geo_types`. It makes it unnecessary to bring to scope the ToWkt trait which gets deleted because it was not being used actually. Of course this is a breaking change.

Additionally I added some docs on the most relevant cases for this crate: parsing a wkt and getting a geometry and having a geometry and getting its WKT string. This is what actually motivates most of the PRs I've been doing, I just wanted to do that .-.

**Some opinions**

What if Wkt was instead a trait itself, implemented for every type in `geo_types` that provided a single method: `wkt(&self) -> String`?

being `georust` an environment for building geo applications it makes sense that things are integrated to work well within the environment, and maybe providing some `LineString` etc. traits which, if implemented on foreign types would give automatically a `Wkt` impl that allows to get its WKT string. And perhaps a `WktGeometry` enum with inner types refering to those in the `geo_types` crate just for the purposes of having something to parse a string into. Although it probably makes more sense to just have `FromStr` impls for the `geo_types` themselves...

Just an idea, I think it would simplify a lot how work with this crate happens.

This actually requires #86 for the docs to pass the tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,11 @@
 # Changes
 
 ## next release
-### Changed
+### Added
 * impl `std::fmt::Display` for `Wkt`.
+* impls of From for the different types in `geo_types` are added.
+### Changed
+* Trait ToWkt gets deprecated.
 
 ## 0.10.0 - 2022-02-24
 ### Changed
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,22 +22,39 @@
 //!
 //! # Examples
 //!
+//! Parse a WKT string to a type:
+//!
 //! ```
-//! use std::str::FromStr;
 //! use wkt::Wkt;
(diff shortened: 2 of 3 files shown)


categulario:
it also addresses #55

michaelkirk:
I'm curious what you though of my proposal in #89 @categulario

categulario:
With `Display` implemented for `Wkt<T>` I don't tink it is necessary because people using `ToWkt` are already getting a `Wkt` out of it, which can be `to_string()`ed

categulario:
Just pushed some commits and rebased main.

ToWkt is restored but deprecated. The doc tests are fixed.

categulario:
Closing this since it no longer reflects my ideas, mostly laid out in #73

PR #94, opened 2022-04-27
Add some benches and some internal cleanup

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [-] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
- [🌮] I ran cargo fmt
---

No changes in behavior, just wanted to do some cleanup and add some additional baseline benchmarks.

PR #89, opened 2022-04-01
Easier serialization of wkt with ToWkt::wkt_string/write_wkt

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
- [x] I ran cargo fmt
---
Fixes #55 - adds an easier API to serialize a geo-type.

~~Depends on #86, so merge that first.~~ Merged!

~~partially conflicts with #88, because I'm relying on the `ToWkt` trait.~~ superseded!

PR #95, opened 2022-04-27
FromWkt trait for reading WKT without exposing the user to the intermediate representation.

michaelkirk:
- [x] I ran cargo fmt
---

Draft because it's based on #94. Please review that first.

--- 

Note that this isn't a performance change. It's about (hopefully) making
the library easier to use.

This is a corollary to https://github.com/georust/wkt/pull/89, but for
reading WKT, rather than writing. As we discussed there, probably there
is no reason for the user to care about the Wkt struct.

Note that the intermediate representation is still used (for now!), but
the user is no longer required to interact with it.

The road is open though for having a direct translation from Wkt text to
the geo-types (or whatever) represenation (see geozero for inspiration).

--- 

I also added the missing GeometryCollection::from(wkt) implementation to make this work for all geo-types.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,8 @@
 * New `FromWkt` trait allows a way to convert from a string or reader directly
   to geo-types, without exposing you to the intermediate `Wkt` structs.
     * <https://github.com/georust/wkt/pull/95>
-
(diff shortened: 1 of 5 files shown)


urschrei (approved):
This is a big usability improvement imo 👏

michaelkirk:
bors r=urschrei

(rebased)

PR #98, opened 2022-05-04
Document all feature, and show feature requirements

michaelkirk:
- [x] I ran cargo fmt
---

This is similar to what we do in geojson and elsewhere. Though I just learned about doc_auto_cfg which makes this way easier by automatically annotating any required items, rather than needing to try to sprinkle a bunch of calls all over like:

    #[cfg_attr(docsrs, doc(cfg(feature = "geo-types")))]
    mod my_geo_types_module;

e.g.
Screen Shot 2022-05-04 at 3 19 26 PM

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+## UNRELEASED
+
+* docs.rs documentation now shows all optional features.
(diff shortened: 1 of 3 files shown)


frewsxcv (approved):

michaelkirk:
bors r=frewsxcv

PR #99, opened 2022-05-05
deserialize specific geo types

michaelkirk:
- [x] I ran cargo fmt
---

The first two commits are just no-ops to move things around. But the final commit is the interesting one. With the exciting bit at https://github.com/georust/wkt/pull/99/files#diff-95bc9d4f4b531a38268ea56c3b4db680f15a6ab9a2bed1b257ad64217f477e5bR61

The new feature is the introduction of `wkt::deserialize_wkt`

```
 let json = r#"[
   { "geometry": "POINT (3.14 42)", "name": "bob's house" },
   { "geometry": "LINESTRING (0 0, 1 1, 3.14 42)", "name": "bob's route home" }
 ]"#;

 #[derive(serde::Deserialize)]
 struct MyGeomRecord {
     #[serde(deserialize_with = "wkt::deserialize_wkt")]
     pub geometry: geo_types::Geometry<f64>,
     pub name: String,
}
```

In the above case, it's pretty much equivalent to using the existing `deserialize_geometry` method, but this new method also works for _any_ of the types that impl `TryFromWkt`, so notably you can deserialize directly to the inner Geometry members when know you're expecting a particular geometry.

```
 let json = r#"{ "geometry": "LINESTRING (0 0, 1 1, 3.14 42)", "name": "bob's route home" }"#;

 #[derive(serde::Deserialize)]
 struct MyGeomRecord {
     #[serde(deserialize_with = "wkt::deserialize_wkt")]
     pub geometry: geo_types::LineString<f64>,
     pub name: String,
}
```


src/deserialize/mod.rs
```diff
@@ -9,6 +13,93 @@ use std::{
     str::FromStr,
 };
 
+#[cfg(feature = "geo-types")]
+pub mod geo_types;
+
+/// Deserializes any type which implements TryFromWkt from a WKT string.
```
urschrei, line 19: What do you think about including a sentence that clarifies that this functionality is intended for working with structured data (I know this is directly implied by the use of `serde`, but that might not be obvious to the beginner or casual Rust user) as opposed to WKT strings?

urschrei (approved):

src/deserialize/mod.rs
michaelkirk: Good point. I've tried to rework the docs to be a bit clearer, as well as re-ordered the doc examples in https://github.com/georust/wkt/pull/99/commits/5129bb9ac336471da8e173a63110a023f94d2f38

I'll merge now, but let me know if you'd like further changes.

michaelkirk:
bors r=urschrei

src/deserialize/mod.rs
urschrei: That was the only thing that leaped out at me, and the updated docs look good!

Issue #102, opened 2022-05-17
Support deserializing integer geometries


michaelkirk:
#101 added support for serializing integer geometries to WKT, but I hit an issue while implementing deserialization for integer types.

My WIP is here:
https://github.com/georust/wkt/tree/mkirk/try-from-non-floats

rmanoka:
Err-ing if the scalar can't be parsed feels like the right thing to do;  agree that the error message could be more helpful.  I don't think there's a huge use-case for this, so we could wait and hear more thoughts.  Thanks for all the effort!

urschrei:
I agree that `Err` is the way to go here – with a more informative error, its resolution can be left up to the library user.

PR #101, opened 2022-05-17
ToWkt for integer types

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
- [x] I ran cargo fmt
---

Add's serialization for integer types.

FIXES https://github.com/georust/wkt/issues/100

In my first attempt, I also tried to add integer support in deserialization (`TryFromWkt`) but I hit a speed bump. I'll open an issue with details momentarily.

PR #103, opened 2022-05-18
TryFromWkt for integer geometries

michaelkirk:
- [x] I ran cargo fmt
---

Fixes https://github.com/georust/wkt/issues/102

~~This is based on #101, which @rmanoka already approved (thanks!), but I'm going to wait just a bit longer to merge it in case someone else wanted to take a look at it.~~
**update:** #101 has been merged, and this is ready for review!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ geo-types = { version = "0.7.4", optional = true }
 num-traits = "0.2"
 serde = { version = "1.0", default-features = false, optional = true }
 thiserror = "1.0.23"
+log = "0.4.17"
 
 [dev-dependencies]
 criterion = { version = "0.2" }
--- a/src/geo_types_from_wkt.rs
+++ b/src/geo_types_from_wkt.rs
@@ -16,14 +16,14 @@
 // limitations under the License.
 
 use crate::types::*;
-use crate::{Geometry, TryFromWkt, Wkt, WktFloat};
+use crate::{Geometry, TryFromWkt, Wkt};
 
 use std::any::type_name;
 use std::convert::{TryFrom, TryInto};
 use std::io::Read;
 use std::str::FromStr;
 
-use geo_types::{coord, CoordFloat};
+use geo_types::{coord, CoordNum};
 use thiserror::Error;
 
 #[derive(Error, Debug)]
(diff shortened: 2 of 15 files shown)


michaelkirk:
#101 has been merged, and this is ready for review

src/geo_types_from_wkt.rs
```diff
@@ -38,15 +38,15 @@ pub enum Error {
     },
     #[error("Wrong number of Geometries: {0}")]
     WrongNumberOfGeometries(usize),
-    #[error("Invalid WKT")]
+    #[error("Invalid WKT: {0}")]
```
michaelkirk, line 41: Note the InvalidWKT now includes the error reason.

Cargo.toml
```diff
@@ -14,11 +14,13 @@ geo-types = { version = "0.7.4", optional = true }
 num-traits = "0.2"
 serde = { version = "1.0", default-features = false, optional = true }
 thiserror = "1.0.23"
+log = "0.4.17"
 
 [dev-dependencies]
 criterion = { version = "0.2" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = "1.0"
+env_logger = "0.9.0"
```
lnicola, line 23: Is this used?

src/tokenizer.rs
```diff
@@ -74,20 +75,30 @@ where
...
-                    Err(_) => None,
+                    Ok(parsed_num) => Token::Number(parsed_num),
+                    Err(_) => {
+                        log::warn!(
+                            "Failed to parse input: '{}' as {}",
+                            &number,
+                            type_name::<T>()
+                        );
+                        return Some(Err(
+                            "Unable to parse input number as the desired output type",
```
lnicola, line 94: This is probably not a number, right?

Actually, we might as well include it in the error instead of pulling `log` along.

Cargo.toml
michaelkirk: I only used it locally. I'll remove it.

src/tokenizer.rs
michaelkirk: It is a `numberlike` as per line 83. In particular to this PR is the case when trying to parse "1.23" as an integer, which will fail with this message.

The reason I logged here rather than building a `String` with the error details to the error is that errors throughout the WKT library are `&'static str`. I could do something different here, but it'd affect multiple places since this error is propagated.

michaelkirk:
Rebased since merging #99

urschrei (approved):

Cargo.toml
michaelkirk: I think this is a stale diff - it was removed in https://github.com/georust/wkt/pull/103/commits/5cdefc65794d1dad9022ab79984ecdcc2c60dc4d

michaelkirk:
bors r=urschrei

PR #106, opened 2022-11-05
Use SPDX license expression

gerritsangel:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
- [x] I ran cargo fmt
---

Uses a SPDX license expression as recommended by [Cargo](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields).

This enables e.g. parsing via the [CycloneDX Rust Cargo Plugin](https://github.com/CycloneDX/cyclonedx-rust-cargo).

PR #107, opened 2022-11-06
Fix some clippy warnings

lnicola:
- [x] I ran cargo fmt
---

### DIFF
--- a/src/geo_types_from_wkt.rs
+++ b/src/geo_types_from_wkt.rs
@@ -995,8 +995,8 @@ mod tests {
         let err = not_a_collection.unwrap_err();
         match err {
             Error::MismatchedGeometry {
-                expected: "geo_types::geometry_collection::GeometryCollection<f64>",
-                found: "geo_types::point::Point<f64>",
+                expected: "geo_types::geometry::geometry_collection::GeometryCollection",
+                found: "geo_types::geometry::point::Point",
             } => {}
             e => panic!("Not the error we expected. Found: {}", e),
         }
@@ -1019,8 +1019,8 @@ mod tests {
         let err = not_actually_a_line_string.unwrap_err();
         match err {
             Error::MismatchedGeometry {
-                expected: "geo_types::line_string::LineString<f64>",
-                found: "geo_types::point::Point<f64>",
+                expected: "geo_types::geometry::line_string::LineString",
+                found: "geo_types::geometry::point::Point",
             } => {}
             e => panic!("Not the error we expected. Found: {}", e),
         }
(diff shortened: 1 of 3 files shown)


lnicola:
bors r+

Self-approving as it's a minor change and it unblocks #106.

src/geo_types_from_wkt.rs
```diff
@@ -995,8 +995,8 @@ mod tests {
         let err = not_a_collection.unwrap_err();
         match err {
             Error::MismatchedGeometry {
-                expected: "geo_types::geometry_collection::GeometryCollection<f64>",
-                found: "geo_types::point::Point<f64>",
+                expected: "geo_types::geometry::geometry_collection::GeometryCollection",
```
lnicola, line 998: From https://doc.rust-lang.org/std/any/fn.type_name.html:

> This is intended for diagnostic use. The exact contents and format of the string returned are not specified, other than being a best-effort description of the type.

In this case, `geo_types::geometry` is just a test module, but the compiler happens to pick it since 1.65. This works on my machine, but I think these tests are pretty broken.

michaelkirk: Do you have a suggestion for making it less broken?

lnicola: Not sure off the top of my head. Maybe we can drop the type name from that error and include an enum?

michaelkirk: When you say "these tests are pretty broken" do you mean only that they are brittle? Or do you mean something else?

lnicola: Yes, sorry, brittle. We could ask if type_name is at least stable on a given toolchain version.

lnicola: I asked this on https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/.E2.9C.94.20How.20stable.20is.20the.20output.20of.20.60std.3A.3Aany.3A.3Atype_name.60.3F.

michaelkirk: For the record, I'd guess that this test failed because of changes made to geo_types in the meanwhile, rather than any change in the rust compiler toolchain.

geo-types now has:
1. a default trait impl `=f64`
2. moved the geometry_collection (private) module into the geometry module 

If you wanted to make this test more robust, you could check for the presence of a substring or something rather than an exact string match, or just delete the test.

lnicola: Ah, I didn't even notice`geo-types` in there. Yes, it makes sense, then. Thanks for following up.

PR #110, opened 2023-09-15
Simplify Wkt data structure

michaelkirk:
- [x] I ran cargo fmt
---

~~This is based on #109, so review that first.~~ Merged!

It's a large breaking change, so I'd appreciate getting at least a couple opinions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,11 +67,10 @@
 //! ```
 //! use std::str::FromStr;
 //! use wkt::Wkt;
-//! use wkt::Geometry;
 //!
 //! let wktls: Wkt<f64> = Wkt::from_str("LINESTRING(10 20, 20 30)").unwrap();
-//! let ls = match wktls.item {
-//!     Geometry::LineString(line_string) => {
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -142,7 +141,7 @@ impl<T> WktFloat for T where T: WktNum + Float {}
 
 #[derive(Clone, Debug, PartialEq)]
 /// All supported WKT geometry [`types`]
-pub enum Geometry<T>
```
michaelkirk, line 145: The crux of the change is this and deleting the old `Wkt`.

frewsxcv (approved):

michaelkirk:
This sat here a long time and I forgot about it. I've rebased! Can I get another review?

michaelkirk:
Well, I got an approval a long time ago and it really seems like no one cares much, so I'm going to merge this before it goes stale again.

kylebarron:
Ah the luck of you merging this right when I was working on ZM support 🥲

PR #113, opened 2024-02-07
Support parse POINTZ, POINTM and POINTZM

ariesdevil:
- [x] I ran cargo fmt
---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,6 +168,35 @@ where
                 let x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens);
                 x.map(|y| y.as_item())
             }
+            w if w.eq_ignore_ascii_case("POINTZ") => {
+                let x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &x.0 {
+                    if coord.z.is_none() {
+                        return Err("POINTZ must have a z-coordinate.");
+                    }
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTM") => {
+                let mut x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &mut x.0 {
+                    if coord.z.is_none() {
+                        return Err("POINTM must have an m-coordinate.");
+                    } else {
+                        coord.m = coord.z.take();
+                    }
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTZM") => {
+                let x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &x.0 {
+                    if coord.z.is_none() || coord.m.is_none() {
+                        return Err("POINTZM must have both a z- and m-coordinate");
+                    }
+                }
+                Ok(x.as_item())
+            }
             w if w.eq_ignore_ascii_case("LINESTRING") || w.eq_ignore_ascii_case("LINEARRING") => {
                 let x = <LineString<T> as FromTokens<T>>::from_tokens_with_parens(tokens);
                 x.map(|y| y.as_item())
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -360,6 +392,49 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_points() {
+        // point(x, y)
+        let wkt = <Wkt<f64>>::from_str("POINT (10 20.1)").ok().unwrap();
+        match wkt.item {
+            Geometry::Point(p) => assert_eq!(
+                format!("{:?}", p),
```
lnicola, line 401: Small nit: using the `Debug` representation is somewhat fragile because it can (and has) changed before. Maybe `matches!` would work here without being too annoying to write?

ariesdevil: I'll try.

src/lib.rs
```diff
@@ -168,6 +168,38 @@ where
...
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTM") => {
+                let mut x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &mut x.0 {
+                    if coord.z.is_none() {
+                        return Err("POINTM must have a linear referencing system.");
+                    } else {
+                        coord.m = coord.z;
```
lnicola, line 186: Nit:

```suggestion
                        coord.m = coord.z.take();
```

src/lib.rs
ariesdevil: Fixed, I keep `POINTM` as it is, so that it is easy to distinguish between "z" and "m".

src/lib.rs
```diff
@@ -168,6 +168,37 @@ where
...
+                        return Err("POINTZ must have a z-coordinate.");
+                    }
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTM") => {
+                let mut x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &mut x.0 {
+                    if coord.z.is_none() {
+                        return Err("POINTM must have a linear referencing system.");
```
lnicola, line 184: I couldn't find this in a spec, but "must have an M coordinate" (or "m-coordinate", but let's be consistent) seems more appropriate?

(same for `POINTZM`)

ariesdevil: From [wikipedia](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry)

> Coordinates for geometries may be 2D (x, y), [3D](https://en.wikipedia.org/wiki/Three-dimensional_space) (x, y, z), 4D (x, y, z, m) with an m value that is part of a [linear referencing system](https://en.wikipedia.org/wiki/Linear_referencing_system) or 2D with an m value (x, y, m).

lnicola:
I guess this will make the conversion to `geo-types` lossy.  I don't think Z/M are available there, so there's not much you can do, just wanted to point that out.

ariesdevil:
> I guess this will make the conversion to `geo-types` lossy. I don't think Z/M are available there, so there's not much you can do, just wanted to point that out.↳

Thanks point it out, I see `Coord` in `geo-types` only support `x` and `y`, I think it needs to be refactored later.

src/lib.rs
```diff
@@ -360,6 +391,31 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_points() {
+        // point(x, y)
+        let wkt = <Wkt<f64>>::from_str("POINT (10 20.1)").ok().unwrap();
+        assert!(matches!(wkt.item, Geometry::Point(_)));
```
lnicola, line 398: Why not

```rust
        assert!(matches!(
            wkt.item,
            Geometry::Point(Point(Some(Coord {
                x: 10.0,
                y: 20.1,
                z: None,
                m: None
            })))
        ));
```

src/lib.rs
lnicola: Yeah, but https://postgis.net/docs/ST_MakePointM.html calls them M coordinates and file:///home/grayshade/Downloads/06-103r4_Implementation_Specification_for_Geographic_Information_-_Simple_feature_access_-_Part_1_Common_Architecture_v1.2.1.pdf says:

> In addition, each coordinate carries an "m" ordinate value that is part of some linear reference system.

> The ordinates x, y and z are spatial, and the ordinate m is a measure.

So I think calling it an "M-coordinate" is a little clearer.

src/lib.rs
lnicola: warning: floating-point types cannot be used in patterns
   --> src/lib.rs:402:20
    |
402 |                 y: 20.1,
    |                    ^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #41620 <https://github.com/rust-lang/rust/issues/41620>


Ugh, I see, okay.

src/lib.rs
ariesdevil: fixed

src/lib.rs
lnicola: Wait, the warning was just removed in https://github.com/rust-lang/rust/pull/116284.

Can you leave the test as.. whatever (even the initial `format!` version) and add a comment to change it to a `matches!` when we bump the MSRV to 1.78? Sorry for the back and forth here!

ariesdevil: So, do we need to change back to string checking?

src/lib.rs
```diff
@@ -168,6 +168,35 @@ where
...
+                        return Err("POINTZ must have a z-coordinate.");
+                    }
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTM") => {
+                let mut x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &mut x.0 {
+                    if coord.z.is_none() {
+                        return Err("POINTM must have a m-coordinate.");
```
lnicola, line 184:
```suggestion
                        return Err("POINTM must have an m-coordinate.");
```

src/lib.rs
```diff
@@ -168,6 +168,35 @@ where
...
+                        coord.m = coord.z.take();
+                    }
+                }
+                Ok(x.as_item())
+            }
+            w if w.eq_ignore_ascii_case("POINTZM") => {
+                let x = <Point<T> as FromTokens<T>>::from_tokens_with_parens(tokens)?;
+                if let Some(coord) = &x.0 {
+                    if coord.z.is_none() || coord.m.is_none() {
+                        return Err("POINTZM must have both z-coordinate and m-coordinate");
```
lnicola, line 195:
```suggestion
                        return Err("POINTZM must have both a z- and m-coordinate");
```

src/lib.rs
lnicola: Yeah, let's do that if it's not too much of a hassle :facepalm:.

ariesdevil:
Fixed, PTAL again.

lnicola (approved):
You didn't add that comment about `matches!`, but whatever :grinning:.

ariesdevil:
> You didn't add that comment about `matches!`, but whatever 😀.↳

Sorry, I forgot it🤔

src/lib.rs
michaelkirk: Is there a reason you can't do something like:
```
assert_eq!(p.x, 10.0)
assert_eq!(p.y, 20.1)
assert_eq!(p.z, None)
assert_eq!(p.m, None)
```

I realize that it's 3 more lines, but comparing to the debug string representation as a way of checking for specific field value equivalency is not something we should get in the habit of doing.

lnicola: I think the debug representation is unlikely to change before we can bump our MSRV to 1.78, so it's not such a big deal anyway.

michaelkirk: Sure it's unlikely, but it did change a few months ago. Treating the debug representation as a stable serialization is a bad idea and we should not start doing it here. I'd hate to see that trend percolate around our code base. Just do it the dumb explicit way.

ariesdevil: Fixed it as your suggested, thx. cc @michaelkirk @urschrei

michaelkirk (approved):
Sorry that this sat here so long!

michaelkirk:
We've misspelled  `POINT Z`, `POINT M`, and `POINTZM`. See https://github.com/georust/wkt/issues/114 

Can you take a look @ariesdevil otherwise I think we should revert this.

PR #115, opened 2024-07-17
Fix parsing Z, M, and ZM WKT strings

kylebarron:
- [x] I ran cargo fmt
---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,13 +83,10 @@ use std::str::FromStr;
 use num_traits::{Float, Num, NumCast};
 
 use crate::tokenizer::{PeekableTokens, Token, Tokens};
-use crate::types::LineString;
-use crate::types::MultiLineString;
-use crate::types::MultiPoint;
-use crate::types::MultiPolygon;
-use crate::types::Point;
-use crate::types::Polygon;
-use crate::types::{Dimension, GeometryCollection};
+use crate::types::{
+    Dimension, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon, Point,
+    Polygon,
+};
 
 mod to_wkt;
 mod tokenizer;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,6 +284,7 @@ fn infer_geom_dimension<T: WktNum + FromStr + Default>(
                     tokens.next().unwrap().unwrap();
                     Ok(Dimension::XYZM)
                 }
+                "EMPTY" => Ok(Dimension::XY),
                 _ => Err("Unexpected word before open paren"),
             },
             // Not a word, e.g. an open paren
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,13 +84,13 @@ use std::str::FromStr;
 use num_traits::{Float, Num, NumCast};
 
 use crate::tokenizer::{PeekableTokens, Token, Tokens};
-use crate::types::GeometryCollection;
 use crate::types::LineString;
 use crate::types::MultiLineString;
 use crate::types::MultiPoint;
 use crate::types::MultiPolygon;
 use crate::types::Point;
 use crate::types::Polygon;
+use crate::types::{Dimension, GeometryCollection};
 
 mod to_wkt;
 mod tokenizer;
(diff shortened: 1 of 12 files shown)


kylebarron:
@michaelkirk since the changes from https://github.com/georust/wkt/pull/110 might be fresh on your mind, do you want to look at the merge conflicts here and/or do you have suggestions for how to resolve them? Or do you want to look at the approach I took here and maybe rebase those onto the latest main?

kylebarron:
Some of the tests here come from the GEOS test suite in https://github.com/libgeos/geos/blob/8e525c825caf314d61a6279a426a59ded115681e/tests/unit/io/WKTReaderTest.cpp

michaelkirk:
> @michaelkirk since the changes from https://github.com/georust/wkt/pull/110 might be fresh on your mind, do you want to look at the merge conflicts here and/or do you have suggestions for how to resolve them? Or do you want to look at the approach I took here and maybe rebase those onto the latest main?

Sorry, I wouldn't have merged that if I knew you were working on this. I'll push up a resolution in a moment the changes are pretty mechanical.

michaelkirk:
I'm seeing some tests failing with your branch. Can you fix those up?

```
failures:

---- tests::empty_items stdout ----
thread 'tests::empty_items' panicked at src/lib.rs:380:63:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::lowercase_point stdout ----
thread 'tests::lowercase_point' panicked at src/lib.rs:395:63:
called `Option::unwrap()` on a `None` value

---- types::multipoint::tests::empty_multipoint stdout ----
thread 'types::multipoint::tests::empty_multipoint' panicked at src/types/multipoint.rs:151:63:
called `Result::unwrap()` on an `Err` value: "Unexpected word before open paren"


failures:
    tests::empty_items
    tests::lowercase_point
```

kylebarron:
> Sorry, I wouldn't have merged that if I knew you were working on this.

No worries! It was just bad luck!

kylebarron:
tests all pass locally now!

kylebarron:
Do you know why CI isn't running on this branch?

michaelkirk:
I do not know why CI isn't running for your branch. Maybe related to the conflicts in the current branch? (working on those now)

michaelkirk:
Ok - conflicts resolved.

src/lib.rs
```diff
@@ -272,21 +243,65 @@ where
...
 }
 
+fn infer_geom_dimension<T: WktNum + FromStr + Default>(
+    tokens: &mut PeekableTokens<T>,
+) -> Result<Dimension, &'static str> {
+    if let Some(Ok(c)) = tokens.peek() {
+        match c {
+            // If we match a word check if it's Z/M/ZM and consume the token from the stream
+            Token::Word(x) => match x.as_str() {
+                "Z" => {
```
michaelkirk, line 253: These should be case insensitive - see our matching elsewhere using `eq_ignore_ascii_case`

src/types/mod.rs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 pub use self::coord::Coord;
+pub use self::dimension::Dimension;
```
michaelkirk, line 16: Does this need to be public? It doesn't look like it's used in any public APIs.

michaelkirk (changes requested):

ariesdevil:
So we now only support `POINT Z`, and no longer support `POINTZ`?

michaelkirk:
That's correct. I don't ever use 3d points, but my understanding is that "POINT Z" is the correct spelling. Please let us know if you have compelling counter examples!

michaelkirk (approved):
Nice work, thanks @kylebarron!

kylebarron:
> So we now only support `POINT Z`, and no longer support `POINTZ`?

I added support for `POINTZ` (and `LINESTRINGZ` etc) in the latest commit. It does make the `match` statement kinda verbose; @michaelkirk what do you think?

michaelkirk:
Sure, since there is apparently some precedent for this spelling, I'm OK with (happy to?) support it. Thanks @kylebarron

kylebarron:
I had missed it but chapter 7 of https://www.ogc.org/standard/sfa/ does define WKT for geometries

michaelkirk (approved):

PR #119, opened 2024-10-06
Infer geometry type and dimension from WKT

### LINKED ISSUE #117: API to infer only geometry type/dimension
In cases like GeoArrow, it's nice to know if all input geometries have a common geometry type.

This would mean an API like

```rust
pub enum GeometryType {
    Point,
    LineString,
    ...,
    PointZ,
    PointM,
    PointZM,
}
```
and then 
```rust
pub fn infer_type(wkt: &str) -> GeometryType {}
```

`infer_type` would only parse up to the first `(` character, using that text to infer the geometry type and dimension.

There's a tradeoff between time spent first parsing the text to get the geometry types and then _again_ to actually parse the geometries, vs memory overhead of parsing all objects first to `Wkt` objects and then inferring what type of array builder to use.

But I figure that scanning a string for the first instance of `(` and matching those first characters should be very fast, especially if no numeric parsing needs to happen for that first stage.

Thoughts?

kylebarron:
- [x] I ran cargo fmt
---

Closes #117

### DIFF
--- a/src/infer_type.rs
+++ b/src/infer_type.rs
@@ -65,19 +65,20 @@ pub fn infer_type(input: &str) -> Result<(GeometryType, Option<Dimension>), Stri
             return Err("Invalid WKT; no '(' character and not EMPTY".to_string());
         }
 
-        if input.starts_with(POINT) {
+        // We use contains instead of starts_with to allow leading whitespace
+        if input.contains(POINT) {
             Ok((GeometryType::Point, None))
-        } else if input.starts_with(LINESTRING) {
+        } else if input.contains(LINESTRING) {
             Ok((GeometryType::LineString, None))
-        } else if input.starts_with(POLYGON) {
+        } else if input.contains(POLYGON) {
             Ok((GeometryType::Polygon, None))
-        } else if input.starts_with(MULTIPOINT) {
+        } else if input.contains(MULTIPOINT) {
             Ok((GeometryType::MultiPoint, None))
-        } else if input.starts_with(MULTILINESTRING) {
+        } else if input.contains(MULTILINESTRING) {
             Ok((GeometryType::MultiLineString, None))
-        } else if input.starts_with(MULTIPOLYGON) {
+        } else if input.contains(MULTIPOLYGON) {
             Ok((GeometryType::MultiPolygon, None))
-        } else if input.starts_with(GEOMETRYCOLLECTION) {
+        } else if input.contains(GEOMETRYCOLLECTION) {
             Ok((GeometryType::GeometryCollection, None))
         } else {
             return Err(format!("Unsupported WKT prefix {}", input));
--- a/src/infer_type.rs
+++ b/src/infer_type.rs
@@ -1,0 +1,131 @@
+use crate::types::{Dimension, GeometryType};
+
+const POINT: &str = "POINT";
+const LINESTRING: &str = "LINESTRING";
+const POLYGON: &str = "POLYGON";
+const MULTIPOINT: &str = "MULTIPOINT";
+const MULTILINESTRING: &str = "MULTILINESTRING";
+const MULTIPOLYGON: &str = "MULTIPOLYGON";
+const GEOMETRYCOLLECTION: &str = "GEOMETRYCOLLECTION";
+
+/// Infer the geometry type and dimension from an input WKT string slice.
+///
+/// An `EMPTY` WKT object will return `None` in place of the dimension.
+///
+/// ```
+/// use wkt::infer_type;
+/// use wkt::types::{Dimension, GeometryType};
+///
+/// assert_eq!(
+///     infer_type("POINT (10 20.1)").unwrap(),
+///     (GeometryType::Point, Some(Dimension::XY))
+/// );
+///
+/// assert_eq!(
+///     infer_type("POINT EMPTY").unwrap(),
+///     (GeometryType::Point, None)
+/// );
+/// ```
+pub fn infer_type(input: &str) -> Result<(GeometryType, Option<Dimension>), String> {
+    if let Some((prefix, _suffix)) = input.split_once("(") {
+        let prefix = prefix.to_uppercase();
+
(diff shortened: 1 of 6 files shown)


src/infer_type.rs
```diff
@@ -0,0 +1,131 @@
...
+        };
+
+        Ok((geom_type, Some(dim)))
+    } else {
+        let input = input.to_uppercase();
+        if !input.contains("EMPTY") {
+            return Err("Invalid WKT; no '(' character and not EMPTY".to_string());
+        }
+
+        if input.starts_with(POINT) {
```
michaelkirk, line 68: input = " POINT EMTPY" 

(leading space?)

src/infer_type.rs
```diff
@@ -0,0 +1,131 @@
...
+            Dimension::XYZ
+        } else if dim_str.contains("M") {
+            Dimension::XYM
+        } else {
+            Dimension::XY
+        };
+
+        Ok((geom_type, Some(dim)))
+    } else {
+        let input = input.to_uppercase();
```
michaelkirk, line 63: It'd be nice to avoid allocation by doing a case insensitive match, rather than copying to a String, but usually these are pretty small, so no biggie. Take it or leave it.

src/infer_type.rs
kylebarron: Yeah... do you think that happens often? Is that valid WKT?

src/infer_type.rs
kylebarron: 1. This is in the block that doesn't contain a `(`, so this should only ever be called on a string as long as `MULTIPOLYGON EMPTY` or so. (Unless the input is invalid)
2. I still agree that I'd prefer not having the allocation. But one of the existing tests was testing the validity of `point empty`, and wasn't sure the best way to handle this. I suppose we could match first on the original input, and if none of those match, cast to uppercase and do it again?

src/infer_type.rs
kylebarron: Would this be better as a `contains`?

michaelkirk: > Is that valid WKT?

Your guess is as good as mine.

But apparently we support parsing it: 

Screenshot 2024-10-07 at 13 06 20

src/infer_type.rs
michaelkirk: I think it's not worth complicating the code much. If it's not a couple word change, then I'd leave it as is.

src/infer_type.rs
kylebarron: In theory [`str::contains`](https://doc.rust-lang.org/std/primitive.str.html#method.contains) should have the same performance as `starts_with` when the match is at the beginning of the string, right?

kylebarron: Switched to contains in [`bcd1250` (#119)](https://github.com/georust/wkt/pull/119/commits/bcd1250b732953cf6edf3d7b8203a0573f662a66)

src/infer_type.rs
kylebarron: Well, actually, this PR currently only supports lower case WKT input in the `EMPTY` case. So we have a test that `point EMPTY` passes, but `point (10, 20)` would currently fail.

I figure both arms should do the same. Should we remove lower case support from the `EMPTY` arm or add it to the primary branch?

michaelkirk: > Well, actually, this PR currently only supports lower case WKT input in the EMPTY case.

I don't understand what you mean. To me it looks like there is already similar "to_uppercase" logic in the other branch.

Since I seem to be lost, rather than talking about the structure of your particular code, I'll explain the behavior that I think makes sense:

In all cases:
- it should probably work with input that our library can parse ( which includes being case insensitive and being resilient to whitespace sprinkled before/after/between tokens)
- it should avoid heap allocation if it doesn't overly complicate the code

kylebarron: Oops you're right, I missed that line 😅 https://github.com/georust/wkt/pull/119/files#diff-14a520b6386f7584931dd3fd2cc7a098cdc7137c83cf69699e115cd74bba13ceR31

michaelkirk:
Unit tests caught a bug with our plan.

michaelkirk:
Quick fix I think is to put the MultiFoo checks before the Foo checks.

michaelkirk:
Or maybe better: use https://doc.rust-lang.org/std/primitive.str.html#method.trim

kylebarron:
> Or maybe better: use [doc.rust-lang.org/std/primitive.str.html#method.trim](https://doc.rust-lang.org/std/primitive.str.html#method.trim)

Oh that looks best! And I assume we only really need [`trim_start`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_start). For some reason I initially thought trimming whitespace would require a new allocation, but makes sense it's not required

src/infer_type.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    #[test]
+    fn test_with_leading_whitespace() {
+        assert_eq!(
+            infer_type(" POINT (10 20.1)").unwrap(),
+            (GeometryType::Point, Some(Dimension::XY))
+        );
+
+        assert_eq!(
+            infer_type("POINT EMPTY").unwrap(),
```
michaelkirk, line 122:
```suggestion
            infer_type(" POINT EMPTY").unwrap(),
```

Assuming this block is trying to test with leading whitespace...

michaelkirk (approved):
LGTM! Feel free to merge after fixing up that last test.

src/infer_type.rs
kylebarron: 🫣

kylebarron:
Ah I see it was merged with a merge commit. Are there standard opinions on whether to use merge commit/squash/rebase for merging for georust repos? On my own repos I only allow squash merging for a cleaner commit history

michaelkirk:
I'm apparently way out on a limb, but I think both have their place.

PR #120, opened 2024-10-08
Switch to ryu for float to string

kylebarron:
- [ ] I ran cargo fmt
---

### DIFF
--- a/src/geo_types_to_wkt.rs
+++ b/src/geo_types_to_wkt.rs
@@ -4,7 +4,7 @@ use crate::types::{
     Coord, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon, Point,
     Polygon,
 };
-use crate::{ToWkt, Wkt};
+use crate::{ToWkt, Wkt, WktNum};
 
 /// # Examples
 /// ```
@@ -13,11 +13,12 @@ use crate::{ToWkt, Wkt};
 ///
 /// let geometry: Geometry<f64> = Geometry::Point(point!(x: 1., y: 2.));
 ///
-/// assert_eq!(geometry.wkt_string(), "POINT(1 2)");
+/// assert_eq!(geometry.wkt_string(), "POINT(1.0 2.0)");
 /// ```
 impl<T> ToWkt<T> for geo_types::Geometry<T>
 where
-    T: CoordNum + std::fmt::Display,
+    T: CoordNum + std::fmt::Display + WktNum + WktNum,
+    Wkt<T>: std::fmt::Display,
 {
     fn to_wkt(&self) -> Wkt<T> {
         match self {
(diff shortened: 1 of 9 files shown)


lnicola:
I'm not convinced this is a good idea. From the rustc issue:

> As I understand it, the fact that short numbers perform worse is an inherent aspect of the algorithm. Dtoa goes left to right and avoids generating too many noise digits, so the performance is better on short numbers as reflected in the table. Ryū generates an exact representation and then goes right to left removing noise digits, so the performance is better on the longest numbers.

> The Ryū benchmarks all focus on a random distribution of numbers, and those have on average a lot of digits.

`dtoa` might be a better choice, then, given that real-world coordinates _should_ have fewer fractional digits?

Also: https://xkcd.com/2170/.

michaelkirk:
Since this is seemingly a straight forward performance proposal, we should be able to measure it.  Can you include any benchmarks using plausible real world data that shows improvement? 

Related: #121 

Note: There were no measurable perf differences for me with the existing write benchmark. It's likely that the existing data we've included in the benchmarks is not representative of real world data, so I'd be happy to include different data in the benchmarks if it's useful to show the improvements here.

That said, if we have to contrive really crazy data to get a measurable perf improvement — e.g. if it's only an improvement for people providing gigabytes of molecular level GPS precision (as an extreme example), and a regression for everybody else, then obviously we shouldn't do it.

src/geo_types_to_wkt.rs
```diff
@@ -42,11 +43,12 @@ where
 ///
 /// let point: Point<f64> = point!(x: 1., y: 2.);
 ///
-/// assert_eq!(point.wkt_string(), "POINT(1 2)");
+/// assert_eq!(point.wkt_string(), "POINT(1.0 2.0)");
```
michaelkirk, line 46: I don't know how to feel about this change in behavior. On the one hand it makes our serialization less efficient which is bad, OTOH it conveys to the reader that this was serialized from floating point, so I guess that's useful.

michaelkirk: In the interest of not leaving ambiguous comments, I'll say I prefer the old version, which serialized efficiently, but I'd be willing to let that small demerit go if there are substantial other wins.

That said, if there's a way to use faster float serialization while still maintaining the integer display for .0, I think that'd be best.

src/types/coord.rs
```diff
@@ -29,22 +29,64 @@ where
...
+    ($t: ident) => {
+        impl fmt::Display for Coord<$t> {
+            fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+                let mut buffer = ryu::Buffer::new();
+                let x = buffer.format(self.x);
+
+                let mut buffer = ryu::Buffer::new();
+                let y = buffer.format(self.y);
+
+                write!(f, "{} {}", x, y)?;
```
lnicola, line 42: Nit: I suspect a sequence of `f.write_str` calls would be a little faster here.

lnicola:
> There were no measurable perf differences for me with the existing write benchmark.

We could set up Criterion, I guess. And it might be worth comparing to `dtoa`, which has the same API.

src/types/coord.rs
michaelkirk: Possibly, it might muddy the waters on perf improvements though. I'd prefer to keep everything except ryu the same in this PR.

michaelkirk:
>> There were no measurable perf differences for me with the existing write benchmark.
>
> We could set up Criterion, I guess. And it might be worth comparing to dtoa, which has the same API.

What do you mean? We're already using criterion AFAICT.

src/geo_types_to_wkt.rs
kylebarron: `ryu` doesn't have _any_ options as far as I can tell. So if we switched to ryu I don't think we'd be able to serialized floats without the `.0`

lnicola: We can still check if the string ends with `.0` and strip that off.

PR #121, opened 2024-10-08
include "write" bench in cargo.toml

michaelkirk:
- [x] I ran cargo fmt
---

I included a write benchmark a while ago, but apparently never added it to Cargo.toml 😅

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,10 @@ default = ["geo-types"]
 name = "parse"
 harness = false
 
+[[bench]]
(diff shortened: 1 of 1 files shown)


lnicola:
Should we add the other two, then?

michaelkirk:
> Should we add the other two, then?

I'm not sure what you mean. Are you confused by the 2 .wkt files in that directory?

lnicola:
Yes!

lnicola (approved):

PR #123, opened 2024-10-15
Implement geo-traits for reading from WKT

kylebarron:
- [x] I ran cargo fmt
---

Ref https://github.com/georust/geo/pull/1157

### DIFF
--- a/src/types/coord.rs
+++ b/src/types/coord.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use geo_traits::PointTrait;
+use geo_traits::CoordTrait;
 
 use crate::tokenizer::{PeekableTokens, Token};
 use crate::types::Dimension;
@@ -98,7 +98,7 @@ where
     }
 }
 
-impl<T: WktNum> PointTrait for Coord<T> {
+impl<T: WktNum> CoordTrait for Coord<T> {
     type T = T;
 
     fn dim(&self) -> geo_traits::Dimensions {
@@ -116,11 +116,6 @@ impl<T: WktNum> PointTrait for Coord<T> {
 
     fn y(&self) -> Self::T {
         self.y
-    }
-
-    fn is_empty(&self) -> bool {
-        // the Coord is never empty; the Point contains an `Option<Coord>` and thus may be empty.
-        false
     }
 
     fn nth_unchecked(&self, n: usize) -> Self::T {
@@ -150,7 +145,7 @@ impl<T: WktNum> PointTrait for Coord<T> {
     }
 }
 
-impl<T: WktNum> PointTrait for &Coord<T> {
+impl<T: WktNum> CoordTrait for &Coord<T> {
     type T = T;
 
     fn dim(&self) -> geo_traits::Dimensions {
@@ -160,11 +155,6 @@ impl<T: WktNum> PointTrait for &Coord<T> {
             (false, true) => geo_traits::Dimensions::Xym,
             (false, false) => geo_traits::Dimensions::Xy,
         }
-    }
-
-    fn is_empty(&self) -> bool {
-        // the Coord is never empty; the Point contains an `Option<Coord>` and thus may be empty.
-        false
     }
 
     fn x(&self) -> Self::T {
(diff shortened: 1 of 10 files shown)


src/types/coord.rs
```diff
@@ -96,6 +98,100 @@ where
...
+
+impl<T: WktNum> PointTrait for &Coord<T> {
+    type T = T;
+
+    fn dim(&self) -> geo_traits::Dimensions {
+        match (self.z.is_some(), self.m.is_some()) {
+            (true, true) => geo_traits::Dimensions::XYZM,
+            (true, false) => geo_traits::Dimensions::XYZ,
+            (false, true) => geo_traits::Dimensions::XYM,
+            (false, false) => geo_traits::Dimensions::XYM,
```
michaelkirk, line 156:
```suggestion
            (false, false) => geo_traits::Dimensions::XY,
```

src/types/geometrycollection.rs
```diff
@@ -84,6 +86,27 @@ where
     }
 }
 
+impl<T: WktNum> GeometryCollectionTrait for GeometryCollection<T> {
+    type T = T;
+    type GeometryType<'a> = &'a Wkt<T> where Self: 'a;
+
+    fn dim(&self) -> geo_traits::Dimensions {
+        if self.0.is_empty() {
```
michaelkirk, line 94: Can this be parsed from the input?

`GEOMETRYCOLLECTION ZM EMPTY`

That might be a larger change though - requiring explicit tracking of the dimensionality of the geometry.

kylebarron: I didn't recall that empty geometries could also have a dimension. Does simple features allow that?

michaelkirk: I guess I'm not sure. I just assumed that this was valid:

`GEOMETRYCOLLECTION Z(POLYGON Z EMPTY, POINT Z(1 2 3))`

Maybe it's not?

kylebarron: Yeah you're right, it is valid: 
image
(I assume shapely/geos implement Simple Features accurately)

kylebarron: I'd say that yes, long-term we should figure out a way to maintain the dimension for empty WKT geometries.

src/types/polygon.rs
```diff
@@ -70,6 +72,58 @@ where
     }
 }
 
+impl<T: WktNum> PolygonTrait for Polygon<T> {
+    type T = T;
+    type RingType<'a> = &'a LineString<T> where Self: 'a;
+
+    fn dim(&self) -> geo_traits::Dimensions {
+        // TODO: infer dimension from empty WKT
```
michaelkirk, line 80: Can you please make an issue describing the work to be done here and any context that we might not remember in a hundred years when someone actually does the work?

src/types/polygon.rs
```diff
@@ -70,6 +72,58 @@ where
...
+    fn exterior(&self) -> Option<Self::RingType<'_>> {
+        self.0.first()
+    }
+
+    fn num_interiors(&self) -> usize {
+        self.0.len().saturating_sub(1)
+    }
+
+    unsafe fn interior_unchecked(&self, i: usize) -> Self::RingType<'_> {
+        &self.0[i + 1]
```
michaelkirk, line 123: Based on my latest understanding after https://github.com/georust/geojson/pull/245/files#r1821255299,

Other than the `CoordTrait.nth_unchecked`, shouldn't these `unchecked` methods be using https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked like https://github.com/georust/geojson/pull/245?

michaelkirk (changes requested):
I'm going to make a PR into geo-traits with the unchecked/or_panic variants so we can stop having the same conversation in every geo-traits repo 🤣

michaelkirk:
(other than the `unchecked` stuff, this looks good!)

src/types/polygon.rs
kylebarron: Made https://github.com/georust/wkt/issues/125

michaelkirk:
Let's release geo-traits-0.2.0 first so we can avoid one sem-ver break.

kylebarron:
I updated this to geo-traits 0.2. Can I get a ✅ ?

michaelkirk (approved):

kylebarron:
Thanks! 🚀

Issue #118, opened 2024-09-25
Use ryu for faster float-to-string conversion


kylebarron:
In https://github.com/geoarrow/geoarrow-rs/pull/788 @b4l prototyped a custom implementation of writing to WKT, and he [said](https://github.com/geoarrow/geoarrow-rs/pull/788#discussion_r1771944782)

> It is up to twice as fast as `wkt` and `geo`.

It doesn't look like this `wkt` crate uses `ryu`, and so I'm curious if that's a large part of why his implementation is faster. https://github.com/dtolnay/ryu seems to be a very popular library, and so it doesn't seem like an unreasonable dependency.

It looks like it should be relatively easy to swap in `ryu` and test if that's significantly faster? It looks like https://github.com/georust/wkt/blob/c3088cdc730c4d643b8535513ddcc3b0efca5e95/src/types/coord.rs#L37-L44 is the main place where floats are converted to strings.

I suppose the primary issue with using ryu is that ryu supports only `f32` and `f64`, while `WktNum` supports a broader range of types, including integers.

Assuming that ryu is indeed quite a bit faster, I think it's worth special casing `f32` and `f64` if we can, assuming that most real world data will be `f64`.

(For maintainability reasons, if geoarrow-rs can use `wkt` directly, that's much more appealing to me than rolling our own).

Any thoughts?

michaelkirk:
I think that sounds really promising. I’d prefer not to lose support for integer geometries though. I wonder if there’s a way to accommodate both?On Sep 24, 2024, at 19:22, Kyle Barron ***@***.***> wrote:﻿
In geoarrow/geoarrow-rs#788 @b4l prototyped a custom implementation of writing to WKT, and he said

It is up to twice as fast as wkt and geo.

It doesn't look like this wkt crate uses ryu, and so I'm curious if that's a large part of why his implementation is faster. https://github.com/dtolnay/ryu seems to be a very popular library, and so it doesn't seem like an unreasonable dependency.
It looks like it should be relatively easy to swap in ryu and test if that's significantly faster? It looks like https://github.com/georust/wkt/blob/c3088cdc730c4d643b8535513ddcc3b0efca5e95/src/types/coord.rs#L37-L44 is the main place where floats are converted to strings.
I suppose the primary issue with using ryu is that ryu supports only f32 and f64, while WktNum supports a broader range of types, including integers.
[...]

PR #124, opened 2024-10-26
Implement geo-traits for writing to WKT & perf improvement

kylebarron:
- [x] I ran cargo fmt
---

### Changes:

This slightly changes the implementation for writing to WKT, partially based on @b4l's PR here https://github.com/geoarrow/geoarrow-rs/pull/788.

While porting the code to use traits, I realized that there was an _awful lot_ of string concatenation being performed. So for example, when writing a Polygon to WKT, **every single coordinate** would be allocated to a separate string, and then the coords of each **ring** would be concatenated and then **all rings** would be concatenated.

https://github.com/georust/wkt/blob/e1d838de2218f410422da28a23c26a7f6329505d/src/types/polygon.rs#L42-L52

This instead writes coordinates _directly_ to the output writer. So perhaps the performance differences in the discussion here https://github.com/georust/wkt/issues/118 were not primarily due to `ryu`.

This PR also implements writing for geo-traits objects, and delegates the `std::fmt::Display` implementation to the underlying geo-traits impl.

Benchmark against `main`:

```
> cargo bench --bench write
Gnuplot not found, using plotters backend
to_string small wkt     time:   [719.98 ns 723.19 ns 726.51 ns]
                        change: [-33.175% -32.566% -32.039%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 17 outliers among 100 measurements (17.00%)
  5 (5.00%) low severe
  3 (3.00%) low mild
  2 (2.00%) high mild
  7 (7.00%) high severe

to_string big wkt       time:   [2.3337 ms 2.3425 ms 2.3524 ms]
                        change: [-44.006% -43.647% -43.297%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

geo: serialize small wkt string
                        time:   [818.73 ns 826.57 ns 836.32 ns]
                        change: [-44.024% -37.443% -31.192%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  7 (7.00%) high mild
  2 (2.00%) high severe

geo: serialize big wkt string
                        time:   [2.4689 ms 2.5142 ms 2.5899 ms]
                        change: [-45.157% -42.370% -39.852%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

geo: write small wkt    time:   [815.24 ns 821.34 ns 831.63 ns]
                        change: [-30.751% -28.806% -27.455%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

geo: write big wkt      time:   [2.4814 ms 2.4877 ms 2.4941 ms]
                        change: [-49.589% -45.273% -42.171%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 23 outliers among 100 measurements (23.00%)
  14 (14.00%) low severe
  4 (4.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
```

I also added another bench for writing the `geo::Geometry` object _directly_ instead of first converting it to a `wkt::Wkt`. It's not 100% apples to oranges because I can't use `std::io::sink()` for a `std::fmt::Write` trait input. But it's still faster 🤷‍♂️ 

```
geo: write small wkt    time:   [809.53 ns 812.46 ns 815.71 ns]
                        change: [-1.9040% -1.1064% -0.5136%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low severe
  4 (4.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe

geo: write big wkt      time:   [2.5784 ms 2.6684 ms 2.8063 ms]
                        change: [+3.6982% +7.2652% +13.562%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe

geo: write small wkt using trait
                        time:   [689.63 ns 692.96 ns 696.94 ns]
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

geo: write big wkt using trait
                        time:   [2.3611 ms 2.3872 ms 2.4283 ms]
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) high mild
  6 (6.00%) high severe
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,14 +11,14 @@ pub enum Error {
     UnknownDimension,
     /// Wrapper around `[std::fmt::Error]`
     #[error(transparent)]
-    FmtError(#[from] std::fmt::Error),
+    FmtError(#[from] core::fmt::Error),
 }
 
 impl From<Error> for fmt::Error {
     fn from(value: Error) -> Self {
         match value {
             Error::FmtError(err) => err,
-            _ => std::fmt::Error,
+            _ => core::fmt::Error,
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,8 +2,8 @@ use core::fmt;
 
 use thiserror::Error;
 
+/// Generic errors for WKT writing and reading
 #[derive(Error, Debug)]
-/// WKT to [`geo_types`] conversions errors
 pub enum Error {
     #[error("Only 2D input is supported when writing Rect to WKT.")]
     RectUnsupportedDimension,
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,8 +5,8 @@ use thiserror::Error;
 #[derive(Error, Debug)]
 /// WKT to [`geo_types`] conversions errors
 pub enum Error {
-    // #[error("The WKT Point was empty, but geo_type::Points cannot be empty")]
-    // RectWriter,
+    #[error("Only 2D input is supported when writing Rect to WKT.")]
+    RectUnsupportedDimension,
     #[error("Only defined dimensions and undefined dimensions of 2, 3, or 4 are supported.")]
     UnknownDimension,
     /// Wrapper around `[std::fmt::Error]`
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,4 @@
-use core::fmt;
+use std::fmt;
 
 use thiserror::Error;
 
@@ -11,14 +11,14 @@ pub enum Error {
     UnknownDimension,
     /// Wrapper around `[std::fmt::Error]`
     #[error(transparent)]
-    FmtError(#[from] core::fmt::Error),
+    FmtError(#[from] std::fmt::Error),
 }
 
 impl From<Error> for fmt::Error {
     fn from(value: Error) -> Self {
         match value {
             Error::FmtError(err) => err,
-            _ => core::fmt::Error,
+            _ => std::fmt::Error,
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,24 @@
+use core::fmt;
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+/// WKT to [`geo_types`] conversions errors
+pub enum Error {
+    // #[error("The WKT Point was empty, but geo_type::Points cannot be empty")]
+    // RectWriter,
+    #[error("Only defined dimensions and undefined dimensions of 2, 3, or 4 are supported.")]
+    UnknownDimension,
+    /// Wrapper around `[std::fmt::Error]`
+    #[error(transparent)]
+    FmtError(#[from] std::fmt::Error),
+}
+
+impl From<Error> for fmt::Error {
+    fn from(value: Error) -> Self {
+        match value {
+            Error::FmtError(err) => err,
+            _ => std::fmt::Error,
+        }
+    }
+}
--- a/src/to_wkt/geo_trait_impl.rs
+++ b/src/to_wkt/geo_trait_impl.rs
@@ -33,9 +33,9 @@ impl From<geo_traits::Dimensions> for PhysicalCoordinateDimension {
     }
 }
 
-pub fn write_point<T: WktNum + fmt::Display, G: PointTrait<T = T>, W: Write>(
+pub fn write_point<T: WktNum + fmt::Display, G: PointTrait<T = T>>(
     g: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = g.dim();
     // Write prefix
@@ -59,9 +59,9 @@ pub fn write_point<T: WktNum + fmt::Display, G: PointTrait<T = T>, W: Write>(
     }
 }
 
-pub fn write_linestring<T: WktNum + fmt::Display, G: LineStringTrait<T = T>, W: Write>(
+pub fn write_linestring<T: WktNum + fmt::Display, G: LineStringTrait<T = T>>(
     linestring: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = linestring.dim();
     // Write prefix
@@ -86,9 +86,9 @@ pub fn write_linestring<T: WktNum + fmt::Display, G: LineStringTrait<T = T>, W: 
     }
 }
 
-pub fn write_polygon<T: WktNum + fmt::Display, G: PolygonTrait<T = T>, W: Write>(
+pub fn write_polygon<T: WktNum + fmt::Display, G: PolygonTrait<T = T>>(
     polygon: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = polygon.dim();
     // Write prefix
@@ -123,9 +123,9 @@ pub fn write_polygon<T: WktNum + fmt::Display, G: PolygonTrait<T = T>, W: Write>
     }
 }
 
-pub fn write_multi_point<T: WktNum + fmt::Display, G: MultiPointTrait<T = T>, W: Write>(
+pub fn write_multi_point<T: WktNum + fmt::Display, G: MultiPointTrait<T = T>>(
     multipoint: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = multipoint.dim();
     // Write prefix
@@ -167,13 +167,9 @@ pub fn write_multi_point<T: WktNum + fmt::Display, G: MultiPointTrait<T = T>, W:
     Ok(())
 }
 
-pub fn write_multi_linestring<
-    T: WktNum + fmt::Display,
-    G: MultiLineStringTrait<T = T>,
-    W: Write,
->(
+pub fn write_multi_linestring<T: WktNum + fmt::Display, G: MultiLineStringTrait<T = T>>(
     multilinestring: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = multilinestring.dim();
     // Write prefix
@@ -209,9 +205,9 @@ pub fn write_multi_linestring<
     Ok(())
 }
 
-pub fn write_multi_polygon<T: WktNum + fmt::Display, G: MultiPolygonTrait<T = T>, W: Write>(
+pub fn write_multi_polygon<T: WktNum + fmt::Display, G: MultiPolygonTrait<T = T>>(
     multipolygon: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = multipolygon.dim();
     // Write prefix
@@ -261,9 +257,9 @@ pub fn write_multi_polygon<T: WktNum + fmt::Display, G: MultiPolygonTrait<T = T>
 
 /// Create geometry to WKT representation.
 
-pub fn write_geometry<T: WktNum + fmt::Display, G: GeometryTrait<T = T>, W: Write>(
+pub fn write_geometry<T: WktNum + fmt::Display, G: GeometryTrait<T = T>>(
     geometry: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     match geometry.as_type() {
         geo_traits::GeometryType::Point(point) => write_point(point, f),
@@ -281,13 +277,9 @@ pub fn write_geometry<T: WktNum + fmt::Display, G: GeometryTrait<T = T>, W: Writ
     }
 }
 
-pub fn write_geometry_collection<
-    T: WktNum + fmt::Display,
-    G: GeometryCollectionTrait<T = T>,
-    W: Write,
->(
+pub fn write_geometry_collection<T: WktNum + fmt::Display, G: GeometryCollectionTrait<T = T>>(
     gc: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     let dim = gc.dim();
     // Write prefix
@@ -322,9 +314,9 @@ pub fn write_geometry_collection<
     Ok(())
 }
 
-pub fn write_rect<T: WktNum + fmt::Display, G: RectTrait<T = T>, W: Write>(
+pub fn write_rect<T: WktNum + fmt::Display, G: RectTrait<T = T>>(
     rect: &G,
-    f: &mut W,
+    f: &mut impl Write,
 ) -> Result<(), Error> {
     // Write prefix and error if not 2D
     match rect.dim() {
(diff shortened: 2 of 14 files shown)


kylebarron:
All existing tests pass; just a question of whether `geo-traits` should be excluded from default features because it depends on `geo-types`, which is excluded.

michaelkirk (changes requested):
In general, can you address the `todo!` before marking a PR as ready for review? I don't want to accidentally miss one and then merge code that willfully explodes on the happy path. It seems like a good way to get users to not trust you.

kylebarron:
Well I consider the `todo` (at least for this pr) "I don't know the right way to handle this, I'm open to advice"

Is there a better way to signify this?

michaelkirk:
Draft PRs are a great way to get line by line feedback and signify "please don't merge this" as is.

I'll also often leave comments on my own PR with specific questions and relevant context at the spot I'm seeking help.

urschrei:
Another alternative / complementary approach is to deny `todo!` in Clippy so a merge becomes gated on no `todo`s

kylebarron:
> Draft PRs are a great way to get line by line feedback and signify "please don't merge this" as is.

Ok! I've found different expectations in different communities on github of whether draft means "don't look at this yet" or "I'm soliciting feedback on things, but it's not yet ready for merge"

kylebarron:
Other questions:

- Should we change the existing `geo_types_to_wkt` behavior? That _clones_ geo-types objects into `wkt` objects, and then writes those. Could we deprecate that entire module?
- This is a PR into the other branch. Any opinions on whether to review and merge that PR first?

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+        match value.size() {
+            2 => Self::Two,
+            3 => Self::Three,
+            4 => Self::Four,
+            size => panic!("Unexpected dimension for coordinate: {}", size),
+        }
+    }
+}
+
+pub fn point_to_wkt<T: CoordNum + WktNum + fmt::Display, G: PointTrait<T = T>, W: Write>(
```
michaelkirk, line 36: naming nit for all the `foo_to_wkt` methods: Idiomatically, I'd expect `point_to_wkt` to do something like `From::from(point) -> Wkt`.

But this doesn't convert to Wkt. Internally it conceptually converts to something like a String of wkt, but we don't return that either.

A closer analogy might be the `Format::write` method or `write` macro, implying something like:
```
write_wkt_point(&mut writer, &my_point)?;
write_wkt_polygon(&mut writer, &my_polygon)?;
```

In fact, I think we can exclude `wkt` from the method names since it's redundant with the crate.

```
write_point(&mut writer, &my_point)?;
write_polygon(&mut writer, &my_polygon)?;
```

If calling users want to be explicit about the fact that it's *wkt* at the callsite, they could:
```
wkt::write_point(&mut writer, &my_point)?;
wkt::write_polygon(&mut writer, &my_point)?;
```

Note: this would entail re-exporting all these methods at the top-level of the crate.


Additionally, taking advantage of https://github.com/georust/geo/pull/1157#issuecomment-2408643696 maybe there should be a plain old:

```
fn write(&mut writer, any_geometry: impl GeometryTrait) {
   write_geometry(writer, any_geometry)
}
```

So the caller can do:
```
wkt::write(&mut writer, &my_point);
wkt::write(&mut writer, &my_polygon);
```

I think that'd work, right? It'd be slightly less performant if you already know the type, but probably negligible for most, and nicer to write.

src/types/coord.rs
```diff
@@ -31,22 +30,6 @@ where
     pub m: Option<T>,
 }
 
-impl<T> fmt::Display for Coord<T>
```
michaelkirk, line 34: Though it seems unlikely anyone is using it, can we leave it as it was and deprecate it first, for sake of avoiding a breaking change?

michaelkirk (changes requested):

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+fn rect_to_polygon<T: CoordNum + WktNum + fmt::Display, G: RectTrait<T = T>>(
+    rect: &G,
+) -> Polygon<T> {
+    let min_coord = rect.min();
+    let max_coord = rect.max();
+
+    // Note: Even if the rect has more than 2 dimensions, we omit the other dimensions when
+    // converting to a Polygon.
```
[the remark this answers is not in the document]
kylebarron: `Triangle` and `Line` store _all_ coordinates, just with the simplifying case that they only have a specific number of points. `Rect` doesn't do that, or else `Rect` would store _4_ coordinates, not 2. We need to do some sort of conversion from the `min` and `max` coordinates to a continuous ring.

It's obvious how to do this for 2D input, but less clear how to do this for 3D or 4D. I'd suggest that we should error for 3D/4D, as the user can choose how to transform a 3D or 4D Rect.

Or perhaps we could even change the `RectTrait` so that it doesn't have the same dimensionality as the others. Because `Rect` means "axis aligned bounding box" and when you have 3 or 4 axes, it's less clear how that maps to simple-features.

src/to_wkt/geo_trait_impl.rs
kylebarron: Yes, this is a good point. I had the same realization when cleaning up my `wkb` crate: https://github.com/kylebarron/wkb/pull/4

src/to_wkt/geo_trait_impl.rs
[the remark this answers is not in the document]
michaelkirk: oh geeze, yeah that's a mess.

Given that we're only handling 2d, can we at least avoid the vec allocation in that case?

As for erroring on 3-d/4-d, that seems reasonable, assuming you mean a runtime error.

src/to_wkt/mod.rs
```diff
@@ -1,5 +1,7 @@
 use crate::{Wkt, WktNum};
 
+pub mod geo_trait_impl;
```
kylebarron, line 3: I don't love the name of this module. Especially if it's becoming the de-facto writer.

src/types/coord.rs
kylebarron: image
is it possible to deprecate a trait impl? Seems not. https://users.rust-lang.org/t/deprecate-an-implementation/90350

src/to_wkt/geo_trait_impl.rs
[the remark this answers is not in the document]
kylebarron: Yeah, we can avoid the vec allocation, just makes the code slightly more complicated.

Also, what should we do with the error? Currently it returns `std::fmt::Error` and there's no error type in this crate.

src/to_wkt/mod.rs
frewsxcv: In case you're copying this naming convention from the `geojson` crate pull request, I went with this name knowing that it wouldn't be exposed publicly. I didn't put a lot of thought into it so feel free to stray from it

src/to_wkt/geo_trait_impl.rs
kylebarron: I shortened names but didn't shorten `write_geometry` to `write`. We should be able to just `pub use write_geometry as write`?

src/to_wkt/mod.rs
kylebarron: I think this really depends on https://github.com/georust/geo/pull/1241 and/or requiring  a `geo-types` dependency here. With this PR _all_ writing goes through the trait impl, but the traits currently depend on `geo-types`.

In that case, we could make `geo_trait_impl` private and re-export the functions through `wkt::to_wkt::*`

kylebarron: ref https://github.com/georust/wkt/pull/124/commits/f3434f5283d070f1accd64cd1bd09303f0f0547c

src/to_wkt/geo_trait_impl.rs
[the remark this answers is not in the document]
michaelkirk: Feel free to propose something.

src/types/coord.rs
michaelkirk: Darn. Sorry for the noise!

kylebarron: Well we could just leave this trait impl in even though it's no longer used

kylebarron:
Let's get #123 merged and then we can come back to this to decide how we want to write `Rect`s

src/to_wkt/geo_trait_impl.rs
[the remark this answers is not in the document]
kylebarron: I added an `Error` enum and now error on non-2d Rect input.

kylebarron:
I updated the Rect writing, changed the `todo!()` to `Error` and this should be good for review now.

src/error.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Wrapper around `[std::fmt::Error]`
+    #[error(transparent)]
+    FmtError(#[from] std::fmt::Error),
+}
+
+impl From<Error> for fmt::Error {
+    fn from(value: Error) -> Self {
+        match value {
+            Error::FmtError(err) => err,
+            _ => std::fmt::Error,
```
michaelkirk, line 21: nit: `std::fmt` here, but `core::fmt` at the top — settle on one.

michaelkirk: If we swallow the error details here, how does the user ever come to see the error details, e.g. "Only defined dimensions and undefined dimensions of 2, 3, or 4 are supported."

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,488 @@
...
+            3 => Self::Three,
+            4 => Self::Four,
+            size => panic!("Unexpected dimension for coordinate: {}", size),
+        }
+    }
+}
+
+pub fn write_point<T: WktNum + fmt::Display, G: PointTrait<T = T>, W: Write>(
+    g: &G,
+    f: &mut W,
```
michaelkirk, line 38:
```suggestion
    f: &mut impl Write
```
It's a little more readable to use `impl Trait` for these kinds of params.

e.g. 
- https://github.com/rust-lang/rust/blob/dff3e7ccd4a18958c938136c4ccdc853fcc86194/src/tools/rust-analyzer/crates/proc-macro-api/src/json.rs#L29
- https://github.com/rust-lang/rust/blob/dff3e7ccd4a18958c938136c4ccdc853fcc86194/compiler/rustc_pattern_analysis/src/rustc/print.rs#L47

style nit #2: I think conventionally the `impl Write` is the first param for free functions, so all together:

```
pub fn write_point<T: WktNum + fmt::Display, G: PointTrait<T = T>>(f: &mut impl Write, g: &G)

// e.g.
write_point(&mut some_writable, &some_point)
```

(similar for other methods in this file)

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,488 @@
...
+        Ok(())
+    } else {
+        Ok(f.write_str(" EMPTY")?)
+    }
+}
+
+pub fn write_linestring<T: WktNum + fmt::Display, G: LineStringTrait<T = T>, W: Write>(
+    linestring: &G,
+    f: &mut W,
+) -> Result<(), crate::error::Error> {
```
michaelkirk, line 65:
```suggestion
) -> Result<(), Error> {
```

Error is already in scope (similar throughout this file)

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    let mut points = multipoint.points();
+
+    // Note: This is largely copied from `add_coord_sequence`, because `multipoint.points()`
+    // yields a sequence of Point, not Coord.
+    if let Some(first_point) = points.next() {
+        f.write_str("((")?;
+
+        // Assume no empty points within this MultiPoint
+        add_coord(&first_point.coord().unwrap(), f, size)?;
```
michaelkirk, line 155:
```suggestion
        write_coord(&first_point.coord().unwrap(), f, size)?;
```

Seems more in line with rest of the API, but given that it's private I don't care that much.

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,488 @@
...
+        geo_traits::Dimensions::Xyzm | geo_traits::Dimensions::Unknown(4) => {
+            f.write_str("MULTILINESTRING ZM")
+        }
+        geo_traits::Dimensions::Unknown(_) => return Err(Error::UnknownDimension),
+    }?;
+    let size = PhysicalCoordinateDimension::from(dim);
+    let mut line_strings = multilinestring.line_strings();
+    if let Some(first_linestring) = line_strings.next() {
+        f.write_str("(")?;
+        add_coord_sequence(first_linestring.coords(), f, size)?;
```
michaelkirk, line 197:
```suggestion
        write_coord_sequence(first_linestring.coords(), f, size)?;
```

Seems more in line with rest of the API, but given that it's private I don't care that much.

michaelkirk (approved):
This looks great! 

I left a bunch of stylistic nitpicks, but I don't feel that strongly about any of them. Apply whatever makes sense to you and merge away.

src/error.rs
kylebarron: Oh, oops. I don't actually know the difference between `std` and `core`

src/to_wkt/geo_trait_impl.rs
kylebarron: > It's a little more readable to use `impl Trait` for these kinds of params.

Happy to switch. Can you clarify what you mean by "these kinds"? Do you just mean "simple bounds without any parameters"? Like you switched the `impl Write` but not the `impl PointTrait`

michaelkirk: I think anything that **can** be written as `impl Trait` probably should be. Maybe there's an exception, but I can't think of one.

src/to_wkt/geo_trait_impl.rs
```diff
@@ -0,0 +1,480 @@
...
+            y: min_coord.y(),
+            z: None,
+            m: None,
+        },
+    ];
+
+    f.write_str("(")?;
+    write_coord_sequence(f, coords.iter(), PhysicalCoordinateDimension::Two)?;
+    Ok(f.write_char(')')?)
+}
```
b4l, line 367: Could be as well be direct like:

```
write!(
        f,
        "POLYGON ({0} {1},{2} {1},{2} {3},{0} {3},{0} {1})",
        min.x(),
        min.y(),
        max.x(),
        max.y(),
 )?;
```

It was not materialized at the time it was copied; sorry for that.

PR #126, opened 2024-11-26
Update `ToWkt::write_wkt` to not buffer string

kylebarron:
- [x] I ran cargo fmt
---

This avoids buffering when writing WKT to `std::io::Write`. It's considerably faster:

```
geo: write small wkt    time:   [570.44 ns 573.74 ns 578.31 ns]
                        change: [-30.487% -29.975% -29.481%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) high mild
  8 (8.00%) high severe

geo: write big wkt      time:   [2.0220 ms 2.0300 ms 2.0405 ms]
                        change: [-19.849% -19.214% -18.603%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe
```

### DIFF
--- a/src/to_wkt/mod.rs
+++ b/src/to_wkt/mod.rs
@@ -8,6 +8,18 @@ pub use geo_trait_impl::{
     write_rect, write_triangle,
 };
 
+/// A wrapper around something that implements std::io::Write to be used with our writer traits,
+/// which require std::fmt::Write
+struct WriterWrapper<W: std::io::Write>(W);
+
+impl<W: std::io::Write> std::fmt::Write for WriterWrapper<W> {
+    fn write_str(&mut self, s: &str) -> std::fmt::Result {
+        // Sadly, this will lose the content of the error when mapping to std::fmt::Error
+        self.0.write(s.as_bytes()).map_err(|_| std::fmt::Error)?;
+        Ok(())
+    }
+}
+
 /// A trait for converting values to WKT
 pub trait ToWkt<T>
 where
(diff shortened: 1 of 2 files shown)


kylebarron:
I updated the PR description. This is actually considerably faster than collecting a string and then writing that to the output buffer.

src/to_wkt/mod.rs
```diff
@@ -8,6 +8,18 @@ pub use geo_trait_impl::{
...
 };
 
+/// A wrapper around something that implements std::io::Write to be used with our writer traits,
+/// which require std::fmt::Write
+struct WriterWrapper<W: std::io::Write>(W);
+
+impl<W: std::io::Write> std::fmt::Write for WriterWrapper<W> {
+    fn write_str(&mut self, s: &str) -> std::fmt::Result {
+        // Sadly, this will lose the content of the error when mapping to std::fmt::Error
+        self.0.write(s.as_bytes()).map_err(|_| std::fmt::Error)?;
```
kylebarron, line 18: This loss of error content is unfortunate but I'm not sure of a good way around this.

kylebarron (commented):
I'd say I'm content with keeping the direct APIs as using `std::fmt::Write` because we know we're writing UTF8 content. Otherwise you'd have to write to a `Vec<u8>` and then parse that back to a `String` (or maybe trust ourselves and use `from_utf8_unchecked`?)

src/to_wkt/mod.rs
michaelkirk: See #129

michaelkirk (approved):

Issue #125, opened 2024-10-30
Store dimensionality in parsed WKT objects


kylebarron:
WKT strings can store a dimensionality even while being `EMPTY`. E.g. these are valid:

```
POINT EMPTY
POINT Z EMPTY
POINT M EMPTY
POINT ZM EMPTY
```

Currently, the parsed representation does not store this. It only stores a coordinate or a sequence of coordinates, depending on the type:
https://github.com/georust/wkt/blob/b2f73990d0f0b3b800b09d7e868291dce4f3d263/src/types/linestring.rs#L23

I think it would be a worthwhile change to also store the dimensionality as part of the geometry. E.g.:
```rs
pub struct LineString<T: WktNum> {
    coords: Vec<Coord<T>>,
    dim: Dimension,
}
```

This would be nice for two reasons:

1. It would solve the empty geometry problem by storing the dimension even when there are no coordinates.
2. It would make it easier to implement the `dim` method for each geometry trait, because it wouldn't have to recurse into the first coordinate.

Unfortunately, this would be a breaking change because the `.0` element of the geometry types is public.

kylebarron:
> Plus, I think for parsing WKT of unknown dimensions

Is there ever `WKT` with unknown dimensions? `POINT (` implies `XY`. If there's 3d data the spec requires `POINT Z` or `POINT M`, right?

michaelkirk:
As an example, what dimensions does this WKT have?

```
let wkt = Wkt::parse(get_wkt_string());
```

kylebarron:
In this PR, dimension is not a generic. The dimension is never known at compile time, but always known at runtime for valid WKT strings.

PR #128, opened 2024-11-26
docs update & bump to 0.12

kylebarron:
- [x] I ran cargo fmt
---

- Update top-level crate documentation with new information about geo-traits
- Add simple docstrings to all public functions
- Remove panic

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+## 0.12.0 - 2024-11-26
+
+* Writing WKT is now up to 50% faster by avoiding extra allocations and writing directly to an underlying buffer.
+* Any `geo_traits` input is now supported when writing to WKT.
+* Implements `geo_traits` on `Wkt` and all structs in `types`. This allows for easier interoperability when reading WKT data to other representations than `geo-types`.
+* BREAKING: removed the `fmt::Display` impl on `wkt::Coord`.
+
 ## 0.11.1 - 2024-10-10
 
 * Add an `infer_type` function to extract only the geometry type and dimension from a WKT string.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "wkt"
 description = "Rust read/write support for well-known text (WKT)"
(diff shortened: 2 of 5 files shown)


michaelkirk:
Should we make the switch to std::fmt::Write before releasing?

kylebarron:
> Should we make the switch to std::fmt::Write before releasing?

We already have `std::fmt::Write`. I was thinking more whether we should switch to `std::io::Write`. But it sounds like no.

kylebarron:
> The main sticking point for me: It seems like while I can stream WKT to a File (and friends) with io::Write, I can't do that with fmt::Write.

I think you can still stream WKT to `io::Write` with a small adapter: https://github.com/georust/wkt/pull/126/files#diff-afc819097d10bcb89581940115867e201017cd2bc9413c58f01b3dd610170f83R39-R49

Every call to `write_str` gets translated into the underlying `io::Write::write`, right? So I don't think there's any buffering still.

It's mostly a little ugly that you lost the content of any `io::Error` because that gets mapped to a `fmt::Error`.

michaelkirk:
> I think you can still stream WKT to io::Write with a small adapter: https://github.com/georust/wkt/pull/126/files#diff-afc819097d10bcb89581940115867e201017cd2bc9413c58f01b3dd610170f83R39-R49

You're right again!

> It's mostly a little ugly that you lost the content of any io::Error because that gets mapped to a fmt::Error.

I might see a way around this... after I wrap up some georust/geo stuff, I'll take a swing at it.

michaelkirk (approved):
LGTM 

Anything else you want to include before releasing @kylebarron or anyone else?

Otherwise, I'll merge and release this tomorrow to give anybody else a final chance to weigh in.

kylebarron:
It looks good to me! https://github.com/georust/wkt/issues/125 would be nice to get to in the medium term, but I don't think it should block a release.

michaelkirk:
released: https://crates.io/crates/wkt/0.12.0
tagged: https://github.com/georust/wkt/releases/tag/0.12.0

PR #133, opened 2025-04-10
geo-types-0.7.16 has new winding order for Rect

michaelkirk:
- [x] I ran cargo fmt
---


kylebarron:
Is that a breaking change for geo-types?

kylebarron (approved):

kylebarron:
I'm going to merge this as it's necessary to fix the CI here regardless

urschrei:
> Is that a breaking change for geo-types?

I don't believe it was tagged as breaking. It was a fix to align implementation with docs (Rect is a Polygon, Polygon shells have CCW order when constructed)

PR #136, opened 2025-04-10
Default to using `f64` for `WktNum` in WKT types

kylebarron:
- [x] I ran cargo fmt
---

Similarly to `geo-types`, defaults to using `f64` as the `WktNum`. Not sure if this is controversial or not?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 * Your change here.
 
+## 0.13.0 - unreleased
+
+* Default to using `f64` for `WktNum` in WKT types
+
 ## 0.12.0 - 2024-11-27
 
 * Writing WKT is now up to 50% faster by avoiding extra allocations and writing directly to an underlying buffer.
--- a/src/geo_types_from_wkt.rs
+++ b/src/geo_types_from_wkt.rs
@@ -434,7 +434,7 @@ mod tests {
 
     #[test]
     fn convert_linestring() {
-        let w_linestring: Wkt<f64> = LineString(vec![
+        let w_linestring: Wkt = LineString(vec![
             Coord {
                 x: 10.,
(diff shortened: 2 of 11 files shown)


urschrei:
I don't think it's controversial at this point FWIW.

kylebarron:
Is there a way to specify `f64` as the default value in `FromStr` so that we don't need to write the extra `Wkt<f64>` in 
```rs
let wktls: Wkt<f64> = Wkt::from_str("LINESTRING(10 20, 20 30)").unwrap();
```

michaelkirk:
I think it's great. Can you add a changelog entry?

kylebarron:
I added a changelog entry.

michaelkirk (approved):

kylebarron:
Thanks for the review!

michaelkirk:
> Is there a way to specify f64 as the default value in FromStr so that we don't need to write the extra Wkt<f64> in

Not that I'm aware of. We have similar default generics in geo/geo-types and I run into the same thing occasionally.

Some examples:

```rust
struct Foo<T: Default = f64> { 
  field: T, 
}

impl<T: Default> Foo<T> {
    fn new_default() -> Self {
        Self { 
            field: T::default(),
        }
    }
}

fn main() {
    // This won't compile - unable to infer generic type
    // let p = Foo::new_default();
    
    // This compiles (I think unsurprisingly)
    let p = Foo::<f64>::new_default();

    // This compiles (I think unsurprisingly)
    let p: Foo<f64> = Foo::new_default();
    
    // This compiles (I think somewhat surprisingly)
    let p: Foo = Foo::new_default();
    
    // This compiles (What even is this?)
    let p = <Foo>::new_default();
}
```

Issue #132, opened 2025-04-10
Extended dimensionality `wkt!` macro


kylebarron:
I'm working to stabilize and harden the core Array types in [geoarrow-array](https://github.com/geoarrow/geoarrow-rs/tree/main/rust/geoarrow-array). GeoArrow supports arrays of dimensionality `XY` but also Z, M, and ZM dimensions. I need to have good test cases with extended dimensionality data.

geoarrow-array supports building data from any source that implements `geo-traits`. So while I could create my own `PointZ`, `LineStringM`, `MultiPolygonZM`, etc (all the combinations). Why not simplify my test harnesses by using a `wkt` macro.

While [`geo_types::wkt` exists](https://github.com/georust/geo/blob/750ed14a2f01c6a047e2ced76ab12261ef46ba5a/geo-types/src/wkt_macro.rs), it only supports 2D data and doesn't support `POINT EMPTY` because `geo_types::Point` can't represent that.

Therefore, I think it makes sense to add a `wkt!` macro here that supports `POINT EMPTY` as well as Z, M, and ZM dimension WKT.

I think this relies on (or at least would be helped by) https://github.com/georust/wkt/issues/125 so that we can faithfully represent `POINT Z EMPTY`.

PR #137, opened 2025-04-10
`wkt!` macro for compile time checking of static WKT text

kylebarron:
- [x] I ran cargo fmt
---

Originally copied from the `wkt-types` implementation but then started copying from #112, since that looked to have nicer separation of concerns.

This is implemented on top of #135 and that should be merged first.

Closes #112  Closes https://github.com/georust/wkt/issues/132


src/wkt_macro.rs
```diff
@@ -0,0 +1,664 @@
...
+        match point {
+            Wkt::Point(p) => {
+                assert_eq!(p.coord.as_ref().unwrap().x(), 1.0);
+                assert_eq!(p.coord.as_ref().unwrap().y(), 2.0);
+                assert_eq!(p.dim, Dimension::XY);
+            }
+            _ => panic!("Expected a Point"),
+        }
+
+        let point = wkt! { POINT Z (1.0 2.0 3.0 4.0) };
```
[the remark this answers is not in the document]
kylebarron:
```rs
        let x = coord_xyz!(1 2 3);
        let x = point!(Z (1 2 3));
```
these both work though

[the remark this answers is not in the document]
kylebarron: It appeared that it was choking on trying to match the Z literal. I was able to work around this by moving the matching on the `Z` to the top-level match

src/wkt_macro.rs
```diff
@@ -0,0 +1,1020 @@
...
+            $crate::wkt_internal!($($wkt)+)
+        }
+    };
+}
+
+#[macro_export(local_inner_macros)]
+#[doc(hidden)]
+macro_rules! wkt_internal {
+    (POINT $tt: tt) => {
+        Wkt::<f64>::Point(point!($tt))
```
michaelkirk, line 39: It'd be nice if this could work for ints too. Is there a reason the macro hardcodes f64?

Is it so that you can do:
```
let point = wkt!(POINT(0.0 0.0))
```

Rather than the more verbose:
```
let point: Wkt = wkt!(POINT(0.0 0.0))
```

kylebarron: I think this is resolved with #136. So we can omit the `<f64>` here and someone can define `let point: Wkt<T>` if they need to.

Do you have any feedback on the design of this PR?

kylebarron: It looks like if we remove the `::<f64>::`, it infers the type from the value, but also fails for `EMPTY`

image

image

image

michaelkirk: Design wise it seems reasonable. It'd be nice if we could pattern out the various Z/M/ZM variants, but also macros are kind of hard. 

I think this will be a nice feature!

kylebarron: > It'd be nice if we could pattern out the various Z/M/ZM variants, but also macros are kind of hard.

Agreed on both counts 🙂.

Despite the verbosity I'll go ahead with fleshing out the rest of the data types.

kylebarron:
I think all the primitive types now work; still working through edge cases with Geometry collections. E.g. this doesn't fail to compile:
image

kylebarron:
Note: This PR doesn't currently handle closing of polygons, like the `geo_types` wkt macro apparently does

michaelkirk:
> Should wkt! { POINT (1 2) } output a Point or a Wkt?
> the thinking being that it's easier to go from concrete type to enum (e.g. via into()) than downcasting from the enum.

For the same reason, I agree!

michaelkirk:
> I wonder if there's an easy way to write a literal that gets parsed as both rust source code and a string literal? Because then we can just use assert_eq to assert that the WKT string parser is equivalent to the macro generator

Seems like it should be doable with a custom test macro that utilizes [`stringify!`](https://doc.rust-lang.org/std/macro.stringify.html)? Unless it *simplifies* this PR, I'd almost rather wait to incorporate it though, because there's already a lot to see here.

kylebarron:
> Note: This PR doesn't currently handle closing of polygons, like the `geo_types` wkt macro apparently does

Upon closer read, I don't think it's `geo_types::wkt!` that handles that; it's the `geo_types::Polygon` _constructor_ that adds an extra coord if needed.

> > I wonder if there's an easy way to write a literal that gets parsed as both rust source code and a string literal? Because then we can just use assert_eq to assert that the WKT string parser is equivalent to the macro generator
> 
> Seems like it should be doable with a custom test macro that utilizes [`stringify!`](https://doc.rust-lang.org/std/macro.stringify.html)? Unless it _simplifies_ this PR, I'd almost rather wait to incorporate it though, because there's already a lot to see here.

💯 

I hadn't been aware of `stringify`. It still might require our literals to be defined _twice_ once in the `stringify!` macro and once in the `wkt!` macro. Agreed we can leave it for the future. I think the current tests are good enough.

kylebarron:
I used this branch last night in https://github.com/geoarrow/geoarrow-rs/pull/1043 and it worked like a charm. See https://github.com/geoarrow/geoarrow-rs/pull/1043/files#diff-cfdac2ccc25441e1e7f36dda38525bb4468123e46129cb6c8ca5d267ff3db0c5

michaelkirk (approved):
This looks great! 

A couple little nits in the tests with some commented out tests - either (preferred) uncomment them, or delete them if that's complicated for some reason.

Thanks for writing all these tests! 

The implementation is a little verbose, and I have some ideas for patterning out the macros, but I can do that in a followup.

PR #141, opened 2025-04-15
Return `Option<Self>` in fallible constructors

kylebarron:
- [ ] I ran cargo fmt
---

Ref https://github.com/georust/wkt/pull/135#discussion_r2045514056


michaelkirk (approved):
Thanks!

PR #149, opened 2025-05-09
Prototype making `geo-traits` an optional dependency; support both 0.2 and 0.3 versions

kylebarron:
- [ ] I ran cargo fmt
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,11 +11,12 @@ edition = "2021"
 
 [dependencies]
 geo-types = { version = "0.7.8", optional = true }
-geo-traits = "0.3"
 num-traits = "0.2"
 serde = { version = "1.0", default-features = false, optional = true }
 thiserror = "1.0.23"
 log = "0.4.17"
+geo-traits_0_2 = { package = "geo-traits", version = "0.2", optional = true }
+geo-traits_0_3 = { package = "geo-traits", version = "0.3", optional = true }
 
 [dev-dependencies]
 criterion = "0.5.1"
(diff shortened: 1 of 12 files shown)


michaelkirk:
> If geo-traits is required, then how do we choose which version to call from write_wkt?

I think you'd need a version of the functions for each version of the trait you're integrating with.

Maybe something like:

```
mod to_wkt {
    #[cfg(feature=geo_traits_0_2)]
    mod  geo_traits_0_2_to_wkt;
    #[cfg(feature=geo_traits_0_2)]
    pub use geo_traits_0_2_to_wkt::*;

    #[cfg(feature=geo_traits_0_3)]
    mod  geo_traits_0_3_to_wkt;
    #[cfg(feature=geo_traits_0_3)]
    pub use geo_traits_0_3_to_wkt::*;
 }
```

Note this violates the convention that features are additive, and seems like it still might be a breaking change? 
To be truly non-breaking, I think I'd actually expect something like 

```
mod to_wkt {
    mod  geo_traits_0_2_to_wkt;
    pub use geo_traits_0_2_to_wkt::*;

    #[cfg(feature=geo_traits_0_3)]
    mod  geo_traits_0_3_to_wkt;
 }
```

And then the caller would have to explicitly call `to_wkt::geo_traits_0_3_to_wkt::write_geometry`

Like I said, given how much work this is, it feels mostly academic at this point.

kylebarron:
> Like I said, given how much work this is, it feels mostly academic at this point.

What do you think is the right path forward here? For my own purposes, one path forward is a `wkb` release with `geo-traits` 0.2, waiting for a wkb release with `geo-traits` 0.3, and then we can wait a similar amount of time here for a new release with `geo-traits` 0.3. All reading and writing in the `wkb` crate is very integrated into `geo-traits`, and so there's no way to make that an optional dependency there.

PR #152, opened 2026-07-05
Bound GeometryCollection nesting depth to prevent stack overflow

LucaCappelletti94:
Deeply nested `GEOMETRYCOLLECTION` WKT caused stack overflow via unbounded recursion.

Bounds nesting at `MAX_DEPTH` (128) and returns a parse error. Also fixes `from_tokens_with_parens` masking inner parse errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,6 +162,9 @@ impl<T> WktNum for T where T: Num + NumCast + PartialOrd + PartialEq + Copy + fm
 pub trait WktFloat: WktNum + Float {}
 impl<T> WktFloat for T where T: WktNum + Float {}
 
+/// Maximum GeometryCollection nesting depth to prevent stack overflow.
+const MAX_DEPTH: usize = 128;
+
 #[derive(Clone, Debug, PartialEq)]
 /// All supported WKT geometry [`types`]
 pub enum Wkt<T: WktNum = f64> {
@@ -199,6 +202,7 @@ where
     fn from_word_and_tokens(
         word: &str,
         tokens: &mut PeekableTokens<T>,
+        depth: usize,
     ) -> Result<Self, &'static str> {
         // Normally Z/M/ZM is separated by a space from the primary WKT word. E.g. `POINT Z`
         // instead of `POINTZ`. However we wish to support both types (in reading). When written
@@ -206,13 +210,14 @@ where
         // matches here.
         match word {
             w if w.eq_ignore_ascii_case("POINT") => {
-                let x = <Point<T> as FromTokens<T>>::from_tokens_with_header(tokens, None);
+                let x = <Point<T> as FromTokens<T>>::from_tokens_with_header(tokens, None, depth);
                 x.map(|y| y.into())
             }
             w if w.eq_ignore_ascii_case("POINTZ") => {
                 let x = <Point<T> as FromTokens<T>>::from_tokens_with_header(
                     tokens,
                     Some(Dimension::XYZ),
+                    depth,
                 );
                 x.map(|y| y.into())
             }
@@ -220,6 +225,7 @@ where
                 let x = <Point<T> as FromTokens<T>>::from_tokens_with_header(
                     tokens,
                     Some(Dimension::XYM),
+                    depth,
                 );
                 x.map(|y| y.into())
             }
@@ -227,17 +233,20 @@ where
                 let x = <Point<T> as FromTokens<T>>::from_tokens_with_header(
                     tokens,
                     Some(Dimension::XYZM),
+                    depth,
                 );
                 x.map(|y| y.into())
             }
             w if w.eq_ignore_ascii_case("LINESTRING") || w.eq_ignore_ascii_case("LINEARRING") => {
-                let x = <LineString<T> as FromTokens<T>>::from_tokens_with_header(tokens, None);
+                let x =
+                    <LineString<T> as FromTokens<T>>::from_tokens_with_header(tokens, None, depth);
                 x.map(|y| y.into())
             }
             w if w.eq_ignore_ascii_case("LINESTRINGZ") => {
                 let x = <LineString<T> as FromTokens<T>>::from_tokens_with_header(
                     tokens,
                     Some(Dimension::XYZ),
+                    depth,
                 );
                 x.map(|y| y.into())
             }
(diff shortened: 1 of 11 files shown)


src/types/coord.rs
```diff
@@ -44,7 +44,11 @@ impl<T> FromTokens<T> for Coord<T>
 where
     T: WktNum + FromStr + Default,
 {
-    fn from_tokens(tokens: &mut PeekableTokens<T>, dim: Dimension) -> Result<Self, &'static str> {
+    fn from_tokens(
+        tokens: &mut PeekableTokens<T>,
+        dim: Dimension,
+        _depth: usize,
```
kylebarron, line 50: For now could we just hard-code a max depth of 128 so that we don't need a public-facing API change?

LucaCappelletti94: Usually the solution is to have an external API method which does not have the argument, and then an internal private trait that has it and is used internally, but that requires a bit more of refactoring. If you prefer it that way I can refactor it in that direction, which by all means is reasonable.

kylebarron: It's hard to imagine real world data having more than 128 nested geometry collections. Could we hard-code a max depth now, and then punt the "perfect" solution to the future in case anyone has real-world data that requires an improved solution?

LucaCappelletti94: I am not sure I understood. The depth parameter is not the maximum depth. It is the way we know at which depth we are, and therefore how we know it is time to stop recursing. We currently do hardcode 128 as the maximum conceivable value, and I believe to be extremely lenient already, likely much less than that is fine.

Correct payloads are extremely unlikely to be that nested. Malevolent ones that want to nuke your server do, and they are surprisingly short.

lnicola:
There's also https://docs.rs/stacker/ which can come in handy when parsing deeply nested inputs.

The fix here feels a bit more invasive than I'd like but I guess we can't help it.

LucaCappelletti94:
In [sqlparser](https://github.com/apache/datafusion-sqlparser-rs) we used https://docs.rs/recursive/latest/recursive/

lnicola:
> https://docs.rs/recursive/latest/recursive/

That's just `stacker` in a slow-to-compile trench coat.

LucaCappelletti94:
I didn't know it was significantly slower to compile, I would argue its value is in the ergonomics, but that's about it.

LucaCappelletti94:
I would propose to close this PR, opening an issue as a reminder, and opening a new PR using `stacker` as you suggest or any other alternative to it. Whatever decision is reached, likely should also be replicated for consistency here: https://github.com/georust/geozero/pull/299

LucaCappelletti94:
@lnicola I built both variants and timed cold release builds on `wkt`, median of 3, `cargo build --release --all-features`.

| variant | wall | user CPU |
|---|---:|---:|
| `main` | 2.40s | 6.23s |
| `stacker` | 3.28s | 10.42s |
| `recursive` | 3.27s | 11.12s |

`recursive` seems a tad bit slower than `stacker`, but nothing I would describe as drastic, and is much less code, so I would propose we go for that one.

Either way we also need an iterative `Drop` on `GeometryCollection`, which the 128 cap hid this because 128 deep drops fit any thread stack, but once the parser can go deeper, dropping the tree recursively overflows on its own.

I will mirror whathever choice is made in georust/geozero#299.

src/lib.rs
```diff
@@ -199,186 +202,206 @@ where
...
-                x.map(|y| y.into())
-            }
-            w if w.eq_ignore_ascii_case("GEOMETRYCOLLECTIONZM") => {
+            w if w.eq_ignore_ascii_case("GEOMETRYCOLLECTION")
+                || w.eq_ignore_ascii_case("GEOMETRYCOLLECTIONZ")
+                || w.eq_ignore_ascii_case("GEOMETRYCOLLECTIONM")
+                || w.eq_ignore_ascii_case("GEOMETRYCOLLECTIONZM") =>
+            {
+                if depth >= MAX_DEPTH {
+                    return Err("Maximum GeometryCollection nesting depth exceeded");
```
CommanderStorm, line 392: The two string returns are not great.
Can we move them to an thiserror instead? This way, a user can match on this..

src/types/coord.rs
CommanderStorm: @kylebarron not public api in the first place, so fine ^^

https://github.com/georust/wkt/blob/85088d9279e5615e3a59e94cd21213d4196bf0e9/src/lib.rs#L770

CommanderStorm (commented):
Mostly LGTM, hardocding is a good approach, nobody has things that are more nested than 128 I think.

Can we improve the error though so an consumer can actually match on the errror, or would this cascade too far?

michaelkirk (commented):
I think it makes sense to do the simple thing that works for 99.999% of people before we worry about introducing a (seemingly?) expensive dependency to nest arbitrarily deep geometry collections.

Though with the current impl, it's unfortunate that the "depth" parameter is being added to so many irrelevant code paths - it only applies to GeometryCollections, right? 

Can this be reasonably re-architected so that the "depth" handling is more localized to GeometryCollections?

src/lib.rs
michaelkirk: It's a valid point, but unreasonable scope creep for this PR IMO.

CommanderStorm: > But unreasonable scope creep for this PR IMO.

Here is the patchset for this 😉 

[0001-refactor-from-string-based-to-structured-error-handl.patch](https://github.com/user-attachments/files/30204855/0001-refactor-from-string-based-to-structured-error-handl.patch)

michaelkirk: Thanks for the patch. Since it's an orthogonal and well contained side quest of more than a few lines, I think it makes sense to rewrite the error handling of this module in a separate PR.

LucaCappelletti94:
Rebased on `main` and reworked so the depth is no longer threaded through geometries that ignore it. `Point`, `Coord`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString` and `MultiPolygon` are now byte-identical to `main`, and the `FromTokens` trait is back to its original signatures.

`GeometryCollection` owns `MAX_DEPTH` and the limit check, and the only place that passes a depth is the `GEOMETRYCOLLECTION` dispatch in `from_word_and_tokens`. Since it is rather quick to do, I will now also prepared the stacked PR for the errors.

michaelkirk (approved):

